PR #67, opened 2020-12-16
ref(*): Updates reqwest to 0.10.10

thomastaylor312:
This allows us to use streams properly and updates all files


technosophos (approved):
YAAAAAAAY

PR #69, opened 2020-12-18
fix(client): Fixes missing HTTP code mapping

thomastaylor312:
Right now the client will return an error if an invoice is created
that already has all of its parcels. This adds the missing HTTP code
mapping to fix that issue.


thomastaylor312:
Added some more test coverage for this

technosophos (approved):

PR #71, opened 2021-01-08
feat(*): Subpathed parcels

thomastaylor312:
This refactors parcel access to occur on the bindle object it is associated
with rather than bare links to parcels. This was a fairly large change as
it had to be reflected both in the storage and client, as well as anything
that consumed them

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -88,10 +88,9 @@ async fn main() -> std::result::Result<(), ClientError> {
         SubCommand::PushFile(push_opts) => {
             let label =
                 generate_label(&push_opts.path, push_opts.name, push_opts.media_type).await?;
-            // TODO: update this to use the from file method once we can update reqwest
             println!("Uploading file {} to server", push_opts.path.display());
             bindle_client
-                .create_parcel(label, tokio::fs::read(push_opts.path).await?)
+                .create_parcel_from_file(push_opts.bindle_id, &label.sha256, push_opts.path)
                 .await?;
             println!("File successfully uploaded");
         }
@@ -102,14 +101,7 @@ async fn main() -> std::result::Result<(), ClientError> {
                 generate_opts.media_type,
             )
             .await?;
-            tokio::fs::OpenOptions::new()
-                .write(true)
-                .create_new(true) // Make sure we aren't overwriting
-                .open(&generate_opts.output)
-                .await?
-                .write_all(&toml::to_vec(&label)?)
-                .await?;
-            println!("Wrote label to {}", generate_opts.output.display());
+            println!("{}", toml::to_string_pretty(&label)?);
         }
     }
 
@@ -150,7 +142,7 @@ async fn generate_label(
 
 async fn get_parcel<C: Cache + Send + Sync + Clone>(cache: C, opts: GetParcel) -> Result<()> {
     let parcel = cache
-        .get_parcel(&opts.sha)
+        .get_parcel(opts.bindle_id, &opts.sha)
         .await
         .map_err(map_storage_error)?;
     let mut file = tokio::fs::OpenOptions::new()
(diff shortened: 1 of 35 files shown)


bin/client/main.rs
```diff
@@ -188,9 +180,16 @@ async fn get_all<C: Cache + Send + Sync + Clone>(cache: C, opts: Get) -> Result<
         .as_ref()
         .unwrap_or(&zero_vec)
         .iter()
-        .map(|p| (p.label.sha256.clone(), cache.clone(), parcels.clone()))
-        .map(|(sha, c, parcels)| async move {
-            match c.get_parcel(&sha).await {
+        .map(|p| {
```
technosophos, line 183: Ooo! Returning the tuple! That's a cool way of doing things!

docs/label-spec.md
```diff
@@ -1,6 +1,6 @@
 # Label Specification
 
-A _label_ (`label.toml`) contains metadata about a box. It is in TOML format, and is considered immutable once stored.
+A _label_ contains metadata about a parcel. It is in TOML format and exists only in the `parcel` array of an invoice
 
 ```toml
 sha256 = 5b992e90b71d5fadab3cd3777230ef370df75f5b...
```
technosophos, line 6: Oops.. I see that this line is not well-formed TOML. Definitely not necessary to fix as part of this PR. I just noticed it because of GitHub's syntax highlighting

docs/protocol-spec.md
```diff
@@ -5,17 +5,16 @@ Bindle uses HTTP/2 with TLS as a transport protocol. All bodies and responses ex
...
-- `/_p/{parcel-id}`: The path to a parcel ID, where `{parcel-id}` is an exact SHA to a parcel.
+- `/_i/{bindle-name}@{parcel-id}`: The path to a Bindle name and parcel ID, where `{parcel-id}` is an exact SHA of a parcel and `{bindle-name}` follows the same rules as outlined above. Parcels can only be accessed if the client has the proper permissions to access the given bindle and, as such, cannot be accessed directly
     - `GET`: Directly fetch a parcel's opaque data.
     - `HEAD`: Send just the headers of a GET request
-- `/_p`
-    - `POST`: Create a parcel if it does not already exist. This may be disallowed. This MUST post both the label and the parcel data with a `multipart/form-data` content type, with the label data as the first part in the request
+    - `POST`: Create a parcel if it does not already exist. This may be disallowed. The data included in the body must have the same SHA as indicated by the `{parcel-id}` and must exist within the invoice
```
technosophos, line 17: I'm feeling like this part of the spec is coming together nicely. Seeing this in writing, I think we made the right choice, and I am glad we got it done this early

examples/client.rs
```diff
@@ -19,17 +19,34 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let second_sha = inv.missing.expect("Should have missing parcels")[0]
+        .sha256
+        .clone();
+
     // Get one of the created invoices
-    println!("Getting invoice 1");
+    println!("Getting invoice 2");
     let inv = bindle_client
-        .get_invoice("enterprise.com/warpcore/1.0.0")
+        .get_invoice("enterprise.com/warpcore/2.0.0")
```
technosophos, line 49: Somehow I missed exactly how bad the punnery was in here. 😆

src/storage/file/mod.rs
```diff
@@ -329,31 +323,26 @@ impl<T: crate::search::Search + Send + Sync> Storage for FileStorage<T> {
             // not be correct.
             out.flush();
             out.seek(std::io::SeekFrom::Start(0)).await?;
-            trace!("Validating data for SHA {}", sha);
-            validate_sha256(&mut out, label.sha256.as_str()).await?;
-            trace!("SHA {} data validated", sha);
+            trace!("Validating data for SHA {}", parcel_id);
+            validate_sha256(&mut out, parcel_id).await?;
+            trace!("SHA {} data validated", parcel_id);
+            // TODO: Should we also validate length? We use it for returning the proper content length
```
technosophos, line 329: I _think_ that we can skip validating content length here because if we know the SHA matches, we know the length must match. Though I guess if the initial length was miscalculated, then this could not match.

technosophos (approved):
Wow... that was a massive amount of work! Looks great, though. Reading through this, I am so glad we made this change early in the dev cycle. Had we waited, it would have been close to impossible!

src/storage/file/mod.rs
thomastaylor312: This is more along making sure the length as indicated by the label matches what was given

bin/client/main.rs
thomastaylor312: Yeah, I finally figured out how to pass these things properly into an `async move`

PR #72, opened 2021-01-08
ref(*): Renames and documents new provider construct

thomastaylor312:
This also entails one more small change to the provider trait


technosophos (approved):
LGTM

PR #73, opened 2021-01-11
feat(*): Adds new proxy provider

thomastaylor312:
This is a simple proxy provider that forwards requests to a bindle server. It
also necessitated changes to the provider API to account for proxying. The
only part of it I am not completely sure about is the `'static` bound I had
to add. I _think_ it makes sense as we need to make sure the stream lives long
enough, but I am not sure if there could be a better way.


technosophos (approved):
LGTM

PR #74, opened 2021-01-11
Changes cache to use proxy and adds CLI tests

thomastaylor312:


thomastaylor312:
This depends on #73 beings merged

technosophos (approved):
LGTM -- that last bug was an interesting one. Yay for good tests!

PR #76, opened 2021-01-16
Sign and verify parcels on an Invoice

technosophos:
This adds parcel signing and verification to the Invoice object.


Cargo.toml
```diff
@@ -64,6 +64,9 @@ log = "0.4.11"
 env_logger = "0.8"
 dirs = { version = "3.0", optional = true }
 mime_guess = { version = "2.0", optional = true }
+rand = "0.7"
+ed25519-dalek = "1.0.1"
```
thomastaylor312, line 68: Nit: Unless there is a need for us to pin to a specific patch version here and with `base64`, I'd drop the patch number

docs/signing-spec.md
```diff
@@ -129,17 +158,69 @@ label.size = 248098
...
+~
 e1706ab0a39ac88094b6d54a3f5cdba41fe5a901
 098fa798779ac88094b6d54a3f5cdba41fe5a901
 5b992e90b71d5fadab3cd3777230ef370df75f5b
 ```
 
+Note that the sequence `\n~\n` is used as a separator to prevent an attempt to conflate a 
+SemVer with the SHA list.
+
+> TODO: Do we need/want a timestamp on the signature?
```
thomastaylor312, line 177: I wonder if we add an optional signing metadata section to support this. It could be really useful information for some people, but I don't think it warrants inclusion within the signing block itself

src/lib.rs
```diff
@@ -104,6 +107,124 @@ impl Invoice {
...
+    ///
+    /// Note that this signature will be invalidated if any parcels are
+    /// added after this signature.
+    ///
+    /// In the current version of the spec, a signature is generated by combining the
+    /// signer's ID, the invoice version, and a list of parcels, and then performing
+    /// a cryptographic signature on those fields. The result is then stored in
+    /// a `[[siganture]]` block on the invoice. Multiple signatures can be attached
+    /// to any invoice.
+    pub fn sign(
```
thomastaylor312, line 141: Non-blocking super nit: With as big as invoice is getting (and its associated types like `SignatureError`, it might be worth moving invoice to its own module like `Id` and then reexporting it at the top level

src/lib.rs
```diff
@@ -104,6 +107,124 @@ impl Invoice {
...
+
+    /// Sign the parcels on the current package.
+    ///
+    /// Note that this signature will be invalidated if any parcels are
+    /// added after this signature.
+    ///
+    /// In the current version of the spec, a signature is generated by combining the
+    /// signer's ID, the invoice version, and a list of parcels, and then performing
+    /// a cryptographic signature on those fields. The result is then stored in
+    /// a `[[siganture]]` block on the invoice. Multiple signatures can be attached
```
thomastaylor312, line 139: Spelling: Signature

src/lib.rs
```diff
@@ -104,6 +107,124 @@ impl Invoice {
...
+    /// - At least one of the signatures is made with a key on the keyring
+    ///
+    /// To verify a signature, the keyring is loaded, and then for each
+    /// `[[signature]]` object on an invoice, the ciphertext is verified.
+    /// The cleartext can be reconstructed from the invoice itself.
+    ///
+    /// Note that the purpose of the keyring is to ensure that we know about the
+    /// entity that claims to have signed the invoice.
+    ///
+    /// If _any_ signature fails, this should be considered a fatal error.
```
thomastaylor312, line 178: I would also document in here that if no signature blocks are present, this will succeed

src/lib.rs
```diff
@@ -209,6 +330,53 @@ pub struct Group {
...
+    pub signature: String,
+    // The public key, encoded as hex chars
+    pub key: String,
+    pub role: SignatureRole,
+}
+
+#[derive(Error, Debug)]
+pub enum SignatureError {
+    #[error("not all signatures can be verified")]
+    Unverified,
```
thomastaylor312, line 353: Nit: Consider including the underlying signature error from the dalek crate so that it is easier to debug why verification failed

src/lib.rs
```diff
@@ -209,6 +330,53 @@ pub struct Group {
...
+    NoKnownKey,
+}
+
+/// The role of a signer in a signature block.
+///
+/// Signatories on a signature must have an associated role, as defined in the
+/// specification.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum SignatureRole {
+    #[serde(rename = "creator")]
```
thomastaylor312, line 372: Nit: You could use the [serde `rename_all = "lowercase"`](https://serde.rs/container-attrs.html#rename_all) attribute here on the struct instead of individually renaming each one

src/lib.rs
```diff
@@ -104,6 +107,124 @@ impl Invoice {
...
         version_compare(self.bindle.id.version(), requirement)
     }
+
+    fn cleartext(&self, by: String, role: SignatureRole) -> String {
+        let id = self.bindle.id.clone();
+        let mut buf = vec![
+            by.clone(),
+            id.name().to_owned(),
+            id.version_string(),
+            format!("{:?}", role).to_lowercase(),
```
thomastaylor312, line 117: Nit: consider implementing `Display` instead on `SignatureRole` so that you can just use `to_string` rather than calling `Debug` (which I think is more expensive)

src/lib.rs
```diff
@@ -432,4 +601,68 @@ mod test {
...
+        let keypair2 = Keypair::generate(&mut rng);
+        let signer_name2 = "Not Matt Butcher <not.matt@example.com>".to_owned();
+
+        // Put one of the two keys on the keyring
+        let keyring = vec![keypair2.public.clone()];
+
+        // Add two signatures
+        invoice
+            .sign(signer_name1, SignatureRole::Creator, keypair1)
+            .expect("sign the parcela");
```
thomastaylor312, line 651: s/parcela/parcel?

src/lib.rs
```diff
@@ -432,4 +601,68 @@ mod test {
...
+        assert_eq!(2, invoice.signature.as_ref().unwrap().len());
+
+        // With the keyring, the signature should work
+        invoice
+            .verify(keyring)
+            .expect("with keys on the keyring, this should pass");
+
+        // This should fail because at least one key must be present in the keyring
+        assert!(invoice.verify(vec![]).is_err());
+    }
```
thomastaylor312, line 667: We probably also need tests for invalid keys and invalid signatures

docs/signing-spec.md
radu-matei: I agree with an optional metadata section, but why would the timestamp be optional?

(Including it could allow for things like purging bindles with signatures older than a particular timestamp.)

src/lib.rs
```diff
@@ -104,6 +107,124 @@ impl Invoice {
...
+    ///
+    /// Note that the purpose of the keyring is to ensure that we know about the
+    /// entity that claims to have signed the invoice.
+    ///
+    /// If _any_ signature fails, this should be considered a fatal error.
+    pub fn verify(&self, keyring: Vec<PublicKey>) -> Result<(), SignatureError> {
+        match self.signature.as_ref() {
+            None => Ok(()),
+            Some(signatures) => {
+                //let mut matches: Vec<PublicKey> = vec![];
```
radu-matei, line 183: Is this comment still needed?

src/lib.rs
```diff
@@ -104,6 +107,124 @@ impl Invoice {
...
+                    // Verify the signature
+                    self.verify_signature(&s, cleartext.as_bytes())?;
+
+                    // See if the public key is known to us
+                    let pubkey =
+                        base64::decode(s.key.clone()).map_err(|_| SignatureError::CorruptKey)?;
+                    let pko = PublicKey::from_bytes(pubkey.as_slice())
+                        .map_err(|_| SignatureError::CorruptKey)?;
+                    if keyring.contains(&pko) {
+                        //matches.push(pko);
```
radu-matei, line 197: Is the comment still needed?

src/lib.rs
radu-matei: (As opposed to the Spanish and Romanian word for "piece of land"?)

docs/signing-spec.md
thomastaylor312: If a bindle _has_ a metadata section, it MUST have a timestamp. Otherwise, it doesn't need one

src/lib.rs
thomastaylor312: Yes, we'll switch it to the english version of "a piece of land"

src/lib.rs
technosophos: Fixed

src/lib.rs
technosophos: Fixed

src/lib.rs
technosophos: Fixed

src/lib.rs
technosophos: Fixed

src/lib.rs
technosophos: Sending the error could disclose information that should not go back to a caller. So we hide that error from potentially going back to the user. Imagine, for example, that it leaks information about the version of an algorithm or a problem with clock skew. An attacker can then use that information to exploit a vulnerability.

It's a weird problem to have, because we do want to know what happened... but only for debugging. We never want this information to be user-visible. Essentially, that's why all of these errors are pretty generic. About all we want to user to know is that a particular operation did not work.

So then there's the matter of logging: Do we trap, log, and then return the masked error? Is logging that data considered safe? Maybe.... For now, I defaulted to "no."

src/lib.rs
technosophos: I didn't realize that worked on Enums!

src/lib.rs
technosophos: fixed in both places.

src/lib.rs
technosophos: One thing we could do is maybe extend it to say _which_ signature failed. That might be helpful. I will look into it and see if there are any security risks with that.

src/lib.rs
technosophos: I read up on this quite a bit before using `Debug`, and the advice was all to use Debug on enums. However, I just switched it to a Display implementation per your request. None of the arguments I could find really bothered to explain why one should prefer `Debug` except that it is more concise. So... meh.

docs/signing-spec.md
technosophos: Alright, so after talking with both of you, I have made Timestamp a _required_ field. I have also made it a u64 to avoid the kind of nonsense we have with Kubernetes and Go's not-quite-RFC-compliant date formatters. A timestamp is easy to parse, as close to unambiguous as anything in our industry gets, and supported across just about everything.

src/lib.rs
technosophos: done

src/lib.rs
technosophos: I have extended the error messages to print the _key_ that failed. So the same piece of information is always used, which makes it easy to go back to the invoice and see which part failed. (Key is shorter than signature, so it seems better to key on.)

thomastaylor312: Yeah, that seems sane

docs/signing-spec.md
```diff
@@ -19,6 +19,52 @@ For example, by signing a parcel's hash, we can in effect certify that _an entit
...
+Examples of proxy signers:
+- CI systems
+- Users who transfer bindles from one Bindle server to another
+- Import/export tools
+
+## Signing and Timestamps
+
+The date and time of a signature can be an important piece of information for auditing.
+
+Both invoice and parcel signatures require an `at` field with a UNIX timestamp expressed as an unsigned 64-bit integer with no leading zeros.
```
thomastaylor312, line 54: Nit: Should this be using a "MUST have an `at` field"?

docs/signing-spec.md
```diff
@@ -19,6 +19,52 @@ For example, by signing a parcel's hash, we can in effect certify that _an entit
...
+The UNIX timestamp offers this:
+
+- It is well defined
+- It is adopted broadly
+- It is an unambiguous format that can easily be represented, serialized, and parsed
+- It is accurate enough for the problem at hand
+
+Longer textual formats are subject to parsing and serialization ambiguities, as well in quirks of implementation across languages.
+
 ## Signing Parcels
```
thomastaylor312, line 68: So just for my understanding, if someone wants to sign parcels, they'll need to do it when they create the invoice, correct? The label is only present within the invoice and not separate as presented below. 

Nit: I'd recommend showing the label inside of the context of an invoice, since labels are no longer standalone

thomastaylor312 (approved):
I assume we'll implement parcel signing next release? Or did you want to handle that this release?

docs/signing-spec.md
technosophos: I am going to have to rewrite the parcels signing stuff. RIght now, it does not make sense because it is based on the older labeled model. And signing it might not be necessary now if we don't have external labels. I'll get this in a follow-up when I can think through it as its own case.

docs/signing-spec.md
technosophos: Probably. I'll get this when I do the rewrite on the parcel section

PR #77, opened 2021-01-19
fix(tests): Makes sure to use tempdir for cache

thomastaylor312:
The CLI integration tests were not using temporary directories for caching
bindles, which led to weird successes in local testing


technosophos (approved):
YAY!

PR #81, opened 2021-01-22
feat(provider): Adds invoice caching to file provider

thomastaylor312:
This will avoid multiple fetches of the invoice from the file system on a
bindle that has a large number of parcels

### DIFF
--- a/src/provider/file/mod.rs
+++ b/src/provider/file/mod.rs
@@ -6,13 +6,16 @@ use std::convert::TryInto;
 use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
+use std::sync::Arc;
 use std::sync::Mutex;
 use std::task::{Context, Poll};
 
+use ::lru::LruCache;
 use log::{debug, error, trace};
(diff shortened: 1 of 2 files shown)


src/provider/file/mod.rs
```diff
@@ -289,6 +311,12 @@ impl<T: crate::search::Search + Send + Sync> Provider for FileProvider<T> {
     {
         debug!("Creating parcel with SHA {}", parcel_id);
 
+        // TODO: Should we check the bindle ID here or stick with the assumption that most terminal
```
thomastaylor312, line 314: @technosophos I'd like your thoughts on this so I can follow up in another PR. I got to thinking while I was working on this and this was the result 😅

technosophos: My got feel is that we want the bindle ID here

thomastaylor312: Ok I’ll do that as a follow up PR to this one to keep it tidy

technosophos (approved):
LGTM

PR #84, opened 2021-02-01
ref(lib): Moves signatures and invoices to their own modules

thomastaylor312:
Each of these types and their implementations were getting pretty big so it made sense to move them to their own modules and reexport them at the top level


src/invoice.rs
```diff
@@ -0,0 +1,416 @@
...
+/// set to `foo/bar` and version set to `1.0.0`.
+///
+/// Most fields on this struct are singular to best represent the specification. There,
+/// fields like `group` and `parcel` are singular due to the conventions of TOML.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct Invoice {
+    pub bindle_version: String,
+    pub yanked: Option<bool>,
+    pub bindle: BindleSpec,
```
technosophos, line 26: This feels strange. The top-level object for the library is Invoice. Now it's in its own module... which is fine... except that all of its children (BindleSpec, Group, etc.) are now in a _parent_ module. That feels inverted to have the children objects defined in the parent module.

technosophos:
I haven't decided what to do on this. It doesn't feel right at all.

It feels like code was arbitrarily pulled into packages just to keep file sizes at some unnamed ideal line count. I find it confusing. There's not a sense of harmony for why Invoice and Signature are children while the other Invoice objects are in a parent modules. 

For example, if I open `lib.rs`, I see a whole much of objects that don't pertain to may daily usage of Bindle. I wouldn't, for example, construct a new Condition in isolation, or a BindleSpec. So lib.rs feels like a catch-all. Meanwhile, if I go open the `invoice` file, I find it importing all of its subordinates from `super::`, which feels reversed.

So if you construct a mental model of where an Invoice's parts are, you have something like:

```
- CHILDREN.rs
    |- PARENT.rs
    |- OTHER_CHILD.rs
```

I'm thinking maybe it would make more sense to move all of the types out of lib.rs. Or to leave _just_ Invoice in lib.rs and move the rest out into submodules. I think I would prefer something that would preserve the conceptual hierarchy and make it easier to find things based on their relation to each other.

I hope that makes sense. Something about it just feels backwards to me.

thomastaylor312:
I had some similar concerns, but I am not quite sure how I want to rework it. We could flip it around and make all the other types exist elsewhere, but that also feels piecemeal. I feel like _something_ needs to move as `lib.rs` is really big and feels like it now has a bunch of logic shoved all over the place

thomastaylor312:
Ok, I redid this after we discussed offline. Should be better organized now

thomastaylor312:
Trying to figure out why the test is failing with malformed TOML. This isn't failing for me locally

PR #85, opened 2021-02-02
fix(provider): Fixes race condition with file provider

thomastaylor312:
The file provider had a possible race condition where 2 people try to create
the same file or if someone tries to create and read at the same time. This
resolves that by writing first to a "part" file and then renaming it. It also
adds documentation to the provider trait that terminal providers should take
care to handle atomic operations or locking


technosophos (approved):
Love it. Nice and elegant.

PR #94, opened 2021-03-08
feat(server): Adds traits and noop types for authn/z

thomastaylor312:
This also glues in the noop types into the bindle server routes, including some additional (and better) routing for parcels and bindles using warp filters

### DIFF
--- a/src/authn/noop/mod.rs
+++ b/src/authn/noop/mod.rs
@@ -2,12 +2,14 @@ use super::Authenticator;
 use crate::authz::noop::Anonymous;
 
 /// An authenticator that simply returns an anonymous user
+#[derive(Clone, Debug)]
 pub struct NoopAuthenticator;
 
+#[async_trait::async_trait]
 impl Authenticator for NoopAuthenticator {
     type Item = Anonymous;
 
-    fn authenticate(&self, _auth_data: &str) -> anyhow::Result<Self::Item> {
+    async fn authenticate(&self, _auth_data: &str) -> anyhow::Result<Self::Item> {
         Ok(Anonymous)
     }
 }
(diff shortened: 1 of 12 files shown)


src/authn/noop/mod.rs
```diff
@@ -0,0 +1,15 @@
+use super::Authenticator;
+use crate::authz::noop::Anonymous;
+
+/// An authenticator that simply returns an anonymous user
+#[derive(Clone, Debug)]
+pub struct NoopAuthenticator;
```
fibonacci1729, line 6: This all looks great! Just a small optionally ignore-able (at your discretion) nit: Maybe instead of Noop, you call it Always or Unconstrained. This might just be me, but a no-op in the context of authentication isn't immediately obvious.

fibonacci1729 (approved):

src/authn/noop/mod.rs
thomastaylor312: This is fair. I'm all for bikeshedding a name if it adds to understanding. I think I'll change it to `Always`

src/authz/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+    /// Returns the identity or username of the authenticated user
+    fn principal(&self) -> String;
+
+    /// Returns the groups the authenticated user is a member of, generally embedded on something
+    /// like a JWT or fetched from an upstream server
+    fn groups(&self) -> Vec<String>;
+}
+
+/// A trait for any system that can authorize any [`Authorizable`](Authorizable) type
+// TODO: Will this need to be async?
```
radu-matei, line 18: Why would `authenticate` be async but `authorize` wouldn't?

radu-matei (approved):
LGTM

src/authz/mod.rs
thomastaylor312: Authenticate for sure will need to contact something upstream to validate a token. Authorize should be checking something in memory. I was keeping this here until I implemented a real authorizer to see what I needed

PR #95, opened 2021-03-11
Define the `approver` role

technosophos:
This defines an `approver` role and goes into more detail about how verification works in Bindle.

### DIFF
--- a/docs/signing-spec.md
+++ b/docs/signing-spec.md
@@ -30,7 +30,7 @@ Asymmetric cryptography is nice for this because a signer can distribute a publi
 ## Signing and Roles
 
 Signers of an invoice MUST have a role. The following roles are defined by this specification.
-Agents SHOULD initiate a warning and MAY initiate a failure if a user's expected role does not match the given role.
+Agents SHOULD initiate a warning and SHOULD initiate a failure if a user's expected role does not match the given role.
 
 - `creator`: Signer asserts that they are the ones who made this bindle
 - `approver`: Signer asserts that they have verified the contents of the bindle
@@ -274,7 +274,7 @@ at = 1611960357
 
 This section is non-normative, describing how implementations MAY choose to behave.
 
-A Bindle may be signed by any number of keys in any number of roles.
+A bindle may be signed by any number of keys in any number of roles.
 During verification, which of these keys must be verified?
 
 The simplest configuration of Bindle would require that the `creator` key be present in the keyring, and the content is then verified against this key's signature. All other keys MAY be verified.
@@ -290,7 +290,7 @@ In this model, the `creator` signature is the only one we care about in verifica
 ### Strategy 2: Authoritative Integrity
 
 This case might be considered weaker than the above, but it illustrates a different way to delegate according to which entity is considered the stronger authority.
-In this case, we say that if the _creator or an approver_ assert that this package is correct, then we accept it.
+In this case, we say that if the _creator or an approver_ asserts that this package is correct, then we accept it.
 
 This setup is useful for a few configurations:
 
--- a/docs/signing-spec.md
+++ b/docs/signing-spec.md
@@ -30,7 +30,7 @@ Asymmetric cryptography is nice for this because a signer can distribute a publi
 ## Signing and Roles
 
 Signers of an invoice MUST have a role. The following roles are defined by this specification.
-Agents SHOULD initiate a warning and MAY initiate a failure if a user's expected role does not match the given roll.
+Agents SHOULD initiate a warning and MAY initiate a failure if a user's expected role does not match the given role.
 
 - `creator`: Signer asserts that they are the ones who made this bindle
 - `approver`: Signer asserts that they have verified the contents of the bindle
@@ -93,8 +93,10 @@ Examples of proxy signers:
 - Import/export tools
 - Proxy servers
 
-While this is implicit in the definitions, we should make it _explicit_ that a Bindle server can use the same key to sign as a host (for bindles it hosts) and as a proxy (for pass-through and cache features).
-But in each case, the key must be marked for the appropriate role.
+While this is implicit in the definitions, we should make it _explicit_ that a Bindle server can use the same key to sign some invoices as a host (for bindles it hosts) and sign others as a proxy (for pass-through and cache features).
+However, it cannot sign the same invoice with the same key acting as both a host and proxy.
+In each case, the key must be marked for the appropriate role.
+
 ## Signing and Timestamps
 
 The date and time of a signature can be an important piece of information for auditing.
@@ -192,12 +194,13 @@ Verification of an invoice includes the following steps:
     a. Extract the public key
     b. Verify that the key has not already been used in another signature block
         - If a key is used to sign the same invoice multiple times, the implementation SHOULD fail
-    c. Verify the signature using the public key and the cleartext d
+    c. Verify the signature using the public key and the cleartext
         - If verification fails for ANY signature block, fail
     d. Locate the key in the keyring
         - If the key is not located, this is not an error
 5. Apply a key trust strategy (See "Strategies of Key Trust" below)
-    - At minimum, the strategy should be that the `creator` signature is done with a known key.
+    - At minimum, the strategy should be that the `creator` signature is signed with a known key (Strategy 1).
+    - For use on public bindle servers, allowing the `creator or approver` strategy (Strategy 2) is preferred.
 
 
 ## Keyrings
@@ -232,9 +235,10 @@ label_signature = "dsaff678..."
 Fields on the `[[key]]` object:
 
 - `label`: A human-readable label that hints what this key is for
-- `roles`: A list of rolls that the user has granted to the key
+- `roles`: A list of roles that the user has granted to the key
 - `key`: The base64-encoded public key for this label
 - `label_signature`: A signature block for the label, to assert that the label is the same one that was intended by the key creator (optional, may be removed)
+
 ## Reading Signatures as Provenance
 
 ```toml
@@ -273,21 +277,15 @@ This section is non-normative, describing how implementations MAY choose to beha
 A Bindle may be signed by any number of keys in any number of roles.
 During verification, which of these keys must be verified?
 
-According to the text above, a creator key MUST be exist in the keyring, and the content must be verified according to this key.
+The simplest configuration of Bindle would require that the `creator` key be present in the keyring, and the content is then verified against this key's signature. All other keys MAY be verified.
 
-> This condition could be eased if we can figure out another way to assert authority of a package. For example we might say "if either the `creator` key is verified or a `approver` key is verified, then the package can be trusted." This model may better reflect systems like NPM, Cargo, or Homebrew.
+This section provides alternative trust strategies that allow different degrees of verification. Implementations may match their validation practices to the security needs of their environment.
 
-But the behavior for all others is that they MAY be verified.
-
-The intention of this is to allows implementations to determine how aggressively to pursue security.
-
-For example, here are three strategy of security. This is not an exhaustive set, but it illustrates the possibilities.
+Five strategies are discussed here. This is not an exhaustive set, but it illustrates the possibilities.
 
 ### Strategy 1: Creative Integrity
 
 In this model, the `creator` signature is the only one we care about in verification. It is as if we are saying, "If the creator says the package I have is the right one, that is all the evidence I need."
-
-It is likely that in the vast majority of cases, this is sufficient.
 
 ### Strategy 2: Authoritative Integrity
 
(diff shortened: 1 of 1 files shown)


technosophos:
I also removed the parcel signing section, as that no longer applies.

docs/signing-spec.md
```diff
@@ -22,22 +30,59 @@ Asymmetric cryptography is nice for this because a signer can distribute a publi
 ## Signing and Roles
 
 Signers of an invoice MUST have a role. The following roles are defined by this specification.
-Agents SHOULD initiate a warning and MAY initiate a failure if another role is encountered.
+Agents SHOULD initiate a warning and MAY initiate a failure if a user's expected role does not match the given roll.
```
thomastaylor312, line 33: s/roll/role

docs/signing-spec.md
```diff
@@ -22,22 +30,59 @@ Asymmetric cryptography is nice for this because a signer can distribute a publi
...
+Verification may form a trust proxy. That is, a client may decide that if the `creator` is unknown, the bindle can still be trusted if one or more of the `approver` keys is known.
+
+### The Host role
+
+The `host` role denotes that the signer is the Bindle server that accepted the invoice from the `creator`.
+
+Bindle servers MUST add a `host` entry to a bindle upon receiving the bindle from a client and verifying the creator key.
 A bindle server MUST NOT sign the same invoice more than once with the same key.
-Clients MAY reject invoices that are signed by unknown hosts, but they also MAY allow them.
+Clients MAY reject invoices that are signed by unknown hosts.
```
thomastaylor312, line 83: 🤔 This makes me wonder if we should have an endpoint on bindle servers that allows you to fetch its private key

docs/signing-spec.md
```diff
@@ -46,7 +91,10 @@ Examples of proxy signers:
 - CI systems
 - Users who transfer bindles from one Bindle server to another
 - Import/export tools
+- Proxy servers
 
+While this is implicit in the definitions, we should make it _explicit_ that a Bindle server can use the same key to sign as a host (for bindles it hosts) and as a proxy (for pass-through and cache features).
```
thomastaylor312, line 96: Up above it says that an invoice cannot be signed by the same key twice. How does that work here?

docs/signing-spec.md
```diff
@@ -217,13 +190,51 @@ Verification of an invoice includes the following steps:
 3. Reconstruct the cleartext block following the signing rules
 4. For each signature block
     a. Extract the public key
-    b. Verify the signature using the public key and the cleartext
+    b. Verify that the key has not already been used in another signature block
+        - If a key is used to sign the same invoice multiple times, the implementation SHOULD fail
+    c. Verify the signature using the public key and the cleartext d
```
thomastaylor312, line 195: There is a trailing `d` here that should be removed

docs/signing-spec.md
```diff
@@ -217,13 +190,51 @@ Verification of an invoice includes the following steps:
...
         - If verification fails for ANY signature block, fail
-    c. Locate the key in the keyring
+    d. Locate the key in the keyring
         - If the key is not located, this is not an error
-5. Verify that at least one signature block used a key that was present in the keyring (4.c)
-    - If none of the signatures were done with a key we know, fail
+5. Apply a key trust strategy (See "Strategies of Key Trust" below)
+    - At minimum, the strategy should be that the `creator` signature is done with a known key.
```
thomastaylor312, line 200: I'd update this to be "At minimum, the strategy should be that the `creator` or `approver` signature is done with a known key."

This more accurately reflects the case outlined above when a bindle can come from an unknown creator but is approved by a known source

docs/signing-spec.md
```diff
@@ -217,13 +190,51 @@ Verification of an invoice includes the following steps:
...
+label = "https://bindle.example.com"
+roles = ["host"]
+key = "cc453q4..."
+label_signature = "dsaff678..."
+```
+
+Fields on the `[[key]]` object:
 
+- `label`: A human-readable label that hints what this key is for
+- `roles`: A list of rolls that the user has granted to the key
```
thomastaylor312, line 235: s/rolls/roles

docs/signing-spec.md
```diff
@@ -242,14 +253,96 @@ by = "GitHub Action [https://github.com/exammple.com/hello_repo]"
...
+ first signed by the creator, then signed by the proxy, and finally signed by the host.
+
+## Strategies of Key Trust
+
+This section is non-normative, describing how implementations MAY choose to behave.
+
+A Bindle may be signed by any number of keys in any number of roles.
+During verification, which of these keys must be verified?
+
+According to the text above, a creator key MUST be exist in the keyring, and the content must be verified according to this key.
```
thomastaylor312, line 276: "a creator or approver key"

docs/signing-spec.md
```diff
@@ -242,14 +253,96 @@ by = "GitHub Action [https://github.com/exammple.com/hello_repo]"
...
+In this strategy, _at least two roles must be verified_.
+The object of this configuration is double-checking for a compromise.
+
+Here are a few scenarios for multiple attestation:
+
+*Creator and CI Pipeline:* Ostensibly, a CI system will build the same package using the same tooling that the creator used.
+Assuming the CI signs its artifacts with the `proxy` role,
+one could require that both `creator` and `proxy` signatures must be present and be verified. This way, if the creator's system is compromised or misconfigured, this would show up in the form of a clash with the proxy signature. While it doesn't identify whether the proxy or the creator was compromised, it shows up as a clash between two distinct keys--and therefore serves as a signpost for where the problem may be.
+
+*Creator and Host:* A similar pairing can be done between the creator role and the host role. In this case, comparing between the host and creator rolls identifies any case where the package was compromised during or after upload to the server.
```
thomastaylor312, line 314: s/rolls/roles

docs/signing-spec.md
```diff
@@ -242,14 +253,96 @@ by = "GitHub Action [https://github.com/exammple.com/hello_repo]"
...
+
+*Creator and approver:* This pairing is a step beyond the "Authoritative Integrity" section above, as it requires that both a creator and a approver can be verified. In other words, for a bindle to be verified, the creator has to certify that they created it, and the approver has to certify that the bindle has been audited.
+
+From here, we could extrapolate to various other combinatorics (e.g. signer, host, and approver must all have verified signatures). But the core idea remains: a conjunction of verifications provides more security than a disjunction.
+
+## Strategy 4: Greedy Verification
+
+This model is the best default model. In it, there is an obligation to verify the `creator` signature, but verification every other signature in the invoice SHOULD be attempted.
```
thomastaylor312, line 322: s/verification/verifying

thomastaylor312 (changes requested):
Overall looks good. There is a little confusion around whether a `creator` is always required or if the actual requirement is "MUST have a `creator` or `approver`".

docs/signing-spec.md
technosophos: Fixed all three

docs/signing-spec.md
technosophos: Private key? Definitely not. But maybe public key. I haven't tackled key management yet because it is really tricky.

docs/signing-spec.md
technosophos: Ah, yeah... I will clarify that I mean it can sign one invoice as a host while another invoice it could sign as a proxy... using the same key for both roles.

docs/signing-spec.md
technosophos: Fixed

docs/signing-spec.md
technosophos: I updated it to point to the two strategies (Creator is strategy 1, creator or approver is strategy 2) later in the doc.

docs/signing-spec.md
thomastaylor312: Oh sorry, that was a total typo. Meant to say public key

docs/signing-spec.md
technosophos: Fixed

docs/signing-spec.md
technosophos: I completely rewrote this section, so it should be clearer. The tl;dr is that the simplest config is `verify the creator`, while another strategy (one that is more fitting to public internet) is `verify the creator or an approver`. I think I got that clearer this time around.

docs/signing-spec.md
```diff
@@ -22,22 +30,59 @@ Asymmetric cryptography is nice for this because a signer can distribute a publi
...
+(In other words, all bindles MUST be signed by a `creator` key before they can be hosted.)
+When a new bindle is submitted, a Bindle server MUST reject an invoice that is signed by a creator's key for which the server does not have an entry in its keyring.
+In other words, the server MUST have the keys for every creator whose Bindle it accepts to host.
+A bindle server MUST NOT sign a bindle with its host key if it does not have the key of the creator in its keyring.
+This does not necessarily preclude a "caching proxy", where the proxy is repeating a Bindle that has been signed by another set of creator and host keys.
+
+> A _caching proxy_ is a Bindle server that takes a user's request for a bindle, finds that bindle remotely, and then caches that bindle's content locally so that it may re-serve that content at a later date. Such servers MAY sign with a proxy key.
+
+A client SHOULD reject an invoice that is not signed by a known key with the `creator` or `approver` role. A non-normative recommendation would be that a client ought to reject any bindle where neither a creator's nor a approver's signature could be verified.
```
thomastaylor312, line 58: Nit: s/a approver/an approver

thomastaylor312 (approved):
I love the changes. This all looks good to me

docs/signing-spec.md
```diff
@@ -22,22 +30,59 @@ Asymmetric cryptography is nice for this because a signer can distribute a publi
 ## Signing and Roles
 
 Signers of an invoice MUST have a role. The following roles are defined by this specification.
-Agents SHOULD initiate a warning and MAY initiate a failure if another role is encountered.
+Agents SHOULD initiate a warning and MAY initiate a failure if a user's expected role does not match the given role.
```
radu-matei, line 33: Should this be stronger and suggest that clients `SHOULD` initiate a failure if the roles don't match?

docs/signing-spec.md
```diff
@@ -242,14 +257,90 @@ by = "GitHub Action [https://github.com/exammple.com/hello_repo]"
...
+at = 1611960357
 ```
+ Because of the timestamps, we can reconstruct the timeline and see that the key was
+ first signed by the creator, then signed by the proxy, and finally signed by the host.
+
+## Strategies of Key Trust
+
+This section is non-normative, describing how implementations MAY choose to behave.
+
+A Bindle may be signed by any number of keys in any number of roles.
```
radu-matei, line 277: Should this be "a bindle"?

docs/signing-spec.md
```diff
@@ -242,14 +257,90 @@ by = "GitHub Action [https://github.com/exammple.com/hello_repo]"
...
+During verification, which of these keys must be verified?
+
+The simplest configuration of Bindle would require that the `creator` key be present in the keyring, and the content is then verified against this key's signature. All other keys MAY be verified.
```
radu-matei, line 280: This states that all other keys may be verified.

Would it make sense to suggest that, if present, all other keys `SHOULD` be verified?

docs/signing-spec.md
```diff
@@ -242,14 +257,90 @@ by = "GitHub Action [https://github.com/exammple.com/hello_repo]"
...
+Five strategies are discussed here. This is not an exhaustive set, but it illustrates the possibilities.
+
+### Strategy 1: Creative Integrity
+
+In this model, the `creator` signature is the only one we care about in verification. It is as if we are saying, "If the creator says the package I have is the right one, that is all the evidence I need."
+
+### Strategy 2: Authoritative Integrity
+
+This case might be considered weaker than the above, but it illustrates a different way to delegate according to which entity is considered the stronger authority.
+In this case, we say that if the _creator or an approver_ assert that this package is correct, then we accept it.
```
radu-matei, line 293: "if the _creator or an approver_ assert" - "asserts"

docs/signing-spec.md
radu-matei: Got to the next section where this is recommended.
Feel free to ignore this comment.

docs/signing-spec.md
technosophos: Good catch

docs/signing-spec.md
technosophos: I am not sure. The issue for me is that the user maintains their own keychain. If they choose to put a key in there as a `approver` key, but that person creates a packages, I do think that the tooling should let the user know... but I don't know whether it should fail.

Now having said that, I think I changed my mind. It does seem like the right thing to do would be to say, "This package was signed with a `creator` key that you only allow to use to `approve`. Stopping. To fix this, edit your keychain."

(I mean, obviously in a more user-friendly way... but the intent is the same)

technosophos: Changed

PR #96, opened 2021-03-24
updated to make yanking more secure

technosophos:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -56,7 +56,7 @@ Here's a compact version of the above:
 ```console
 $ edit invoice.toml
 $ cat invoice.toml
-bindle_version = "1.0.0"
+bindleVersion = "1.0.0"
 
 [bindle]
 name = "mybindle"
@@ -76,7 +76,7 @@ You can verify that this is working by fetching the invoice:
 ```console
 $ bindle info mybindle/0.1.0
 # request for mybindle/0.1.0
-bindle_version = "1.0.0"
+bindleVersion = "1.0.0"
 
 [bindle]
 name = "mybindle"
--- a/docs/README.md
+++ b/docs/README.md
@@ -56,7 +56,7 @@ Here's a compact version of the above:
 ```console
 $ edit invoice.toml
 $ cat invoice.toml
-bindleVersion = "1.0.0"
+bindle_version = "1.0.0"
 
 [bindle]
 name = "mybindle"
@@ -76,7 +76,7 @@ You can verify that this is working by fetching the invoice:
 ```console
 $ bindle info mybindle/0.1.0
 # request for mybindle/0.1.0
-bindleVersion = "1.0.0"
+bindle_version = "1.0.0"
 
 [bindle]
 name = "mybindle"
--- a/docs/invoice-spec.md
+++ b/docs/invoice-spec.md
@@ -5,7 +5,7 @@ This specification describes the invoice (`invoice.toml`).
 An invoice is the top-level descriptor of a bindle. Every bindle has exactly one invoice.
 
 ```toml
-bindleVersion = "v1.0.0"
+bindle_version = "v1.0.0"
 
 [bindle]
 name = "mybindle"
@@ -41,7 +41,7 @@ The above bindle declares its `bindle` description, and then declares a manifest
 
 ## Top-level Fields
 
-- `bindleVersion` is required, and should be `1.0.0` for this version of the specification.
+- `bindle_version` is required, and should be `1.0.0` for this version of the specification.
 - `yanked` is a boolean field that indicates whether a Bindle has been yanked. This field appears outside of the `bindle` because it is mutable, though it can only be toggled on. Once set to true, a Bindle MUST NOT be un-yanked. A yanked bundle should never be served in an index or search, but MAY be accessed directly.
 - `yanked_reason` (OPTIONAL) is a string field in which a human-readable reason can be given for yanking the invoice.
 
@@ -92,7 +92,7 @@ For any explicitly created group, it is empty by default. Parcels must be placed
 The `[[group]]` list is used to create a group. In the following example, three groups are defined: `server`, `cli`, and `utility`.
 
 ```toml
-bindleVersion = "1.0.0"
+bindle_version = "1.0.0"
 
 [bindle]
 name = "mybindle"
@@ -147,7 +147,7 @@ By default, if no condition is provided, an item is a member of the "global" gro
 Example:
 
 ```toml
-bindleVersion = "1.0.0"
+bindle_version = "1.0.0"
 
 [bindle]
 name = "mybindle"
--- a/docs/invoice-spec.md
+++ b/docs/invoice-spec.md
@@ -43,6 +43,7 @@ The above bindle declares its `bindle` description, and then declares a manifest
 
 - `bindleVersion` is required, and should be `1.0.0` for this version of the specification.
 - `yanked` is a boolean field that indicates whether a Bindle has been yanked. This field appears outside of the `bindle` because it is mutable, though it can only be toggled on. Once set to true, a Bindle MUST NOT be un-yanked. A yanked bundle should never be served in an index or search, but MAY be accessed directly.
+- `yanked_reason` (OPTIONAL) is a string field in which a human-readable reason can be given for yanking the invoice.
 
 ## `bindle` Fields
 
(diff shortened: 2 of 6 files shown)


docs/invoice-spec.md
```diff
@@ -43,6 +43,7 @@ The above bindle declares its `bindle` description, and then declares a manifest
 
 - `bindleVersion` is required, and should be `1.0.0` for this version of the specification.
 - `yanked` is a boolean field that indicates whether a Bindle has been yanked. This field appears outside of the `bindle` because it is mutable, though it can only be toggled on. Once set to true, a Bindle MUST NOT be un-yanked. A yanked bundle should never be served in an index or search, but MAY be accessed directly.
+- `yanked_reason` (OPTIONAL) is a string field in which a human-readable reason can be given for yanking the invoice.
```
radu-matei, line 46: There is inconsistency between the cases for `bindleVersion` and `yanked_reason`.

docs/signing-spec.md
```diff
@@ -342,6 +343,236 @@ For example, if a package is signed by an unknown host key, then this package ha
...
+Note that this strategy does not prevent a key from being used to generate a known-bad signature so that the validation fails.
+
+### Known Issues with Yank Signing
+
+The current model does not provide a way to say "this package should never be trusted again" except for generating known-bad signatures.
+
+Likewise, while signing with `host` keys indicates that the host has yanked the package, there is no clear way to indicate the creator's intent, or a verifier's stance. The reason for this is that yanking is a host-specific operation, and does not have
+to do directly with creator intent. (In other words, the creator is not the authority on whether a bindle is yanked. The host is.)
+
+A compromised host could yank all packages, which may have dire initial consequences.
```
radu-matei, line 411: Is there the possibility of a compromised host un-yanking packages?

docs/invoice-spec.md
technosophos: Oh yeah! we changed `bindleVersion` to `bindle_version` a few PRs ago and I must have missed several spots in the docs. Thanks.

docs/signing-spec.md
technosophos: Oh! That is a very good point. That could be bad.

docs/invoice-spec.md
technosophos: Fixed throughout docs.

docs/signing-spec.md
```diff
@@ -342,6 +343,236 @@ For example, if a package is signed by an unknown host key, then this package ha
...
+
+The `yanked` field could be included in the main signing metadata.
+This would create a _strong yank_ under which the yank would invalidate all signatures.
+
+The main flaw with this route is that it makes a strong claim about what `yanked` means.
+In this case, `yanked` means "it is never secure to use this package", which may be true in some cases, but not all.
+
+Another issue with this approach is that a yanked bindle's non-host signatures would all be initially invalidated by the
+yank operation, even though the actual content of the package has not changed.
+This would require verifiers and creators to go _re-sign_ packages again after the `yanked` field was toggled to true.
```
thomastaylor312, line 424: This could also lead to all sorts of other problems too right? Someone else could claim to be a creator when they aren't. I actually think we may want to specify that signatures are append only

docs/signing-spec.md
```diff
@@ -342,6 +343,236 @@ For example, if a package is signed by an unknown host key, then this package ha
...
+Thus, on yank, a bindle would become unusable until the signers re-signed.
+That was not the intent of the `yanked` field or the yanking operation.
+
+## Threat Model
+
+This section describes our threat model in applying signatures.
+Our inspiration has been [TUF]().
```
thomastaylor312, line 431: Missing link?

docs/signing-spec.md
```diff
@@ -342,6 +343,236 @@ For example, if a package is signed by an unknown host key, then this package ha
...
+
+- No single key holder can be assumed to be trustworthy
+    - A creator is generally an authority on what they wish to produce, but is not necessarily to be trusted to produce non-malicious bindles
+    - An approver, likewise, may produce results that are incomplete or potentially malicious
+    - A proxy or host may be malicious
+- An attacker can act as a man-in-the-middle (MITM) via hijacking or proxying. Therefore, the security model must assume a potential MITM.
+- If an attack can compromise enough keys, the best that can be done by this spec is to limit the distribution of compromised bindles
+- A malicious signer can cause a package's trust to be doubted (e.g. denial of service), but it should not be able to make a doubted package's trust _accepted_.
+    - That is, if other signatures are invalid, a malicious signature should not be able to override that invalidation
+    - But as a weakness in the system, if a malicious signer creates a bad signature, that signature may result in an uncompromised package becoming untrusted for any user who has the malicious signer's signature in their keyring
```
thomastaylor312, line 454: This reads a little confusing to me. Are you saying that if a person has the key of the malicious signer in their local keyring it becomes untrusted? Maybe I'm not reading it right

radu-matei (approved):

docs/signing-spec.md
```diff
@@ -342,6 +343,236 @@ For example, if a package is signed by an unknown host key, then this package ha
...
+An implementation may only mind invalid signatures on keys that are in its keyring or that are in a certain role.
+
+## Key Rotation
+
+Bindle makes no assumptions about the frequency of key rotation.
+However, there are a few non-normative recommendations:
+
+- If a `host` key is rotated, all packages SHOULD be re-signed by the new host key.
+    - For a properly configured Bindle server, this does not introduce an opportunity for an attacker to modify packages because the `creator` and all `proxy` and `approver` signatures remain unaltered, and any attempt to modify the package will invalidate those signatures
+- When a creator rotates keys, and the cause is NOT compromise, the user SHOULD re-sign existing bindles using the new key and the `approver` role. The original `creator` signature SHOULD NOT be altered in any way.
```
thomastaylor312, line 560: I could be misremembering, but isn't a bindle only allowed to have one `creator` signature? If I am a consumer, and I see two creator signatures, what does that mean? I can see someone else asserting they are a creator when they aren't

thomastaylor312 (changes requested):

docs/signing-spec.md
```diff
@@ -342,6 +343,236 @@ For example, if a package is signed by an unknown host key, then this package ha
...
+To verify a `yanked_signature`, an agent:
+
+- MUST verify that the host key used to sign is in the agent's keyring
+    - An agent MAY refuse to treat a bindle as yanked if the key is not known
+    - An agent SHOULD notify the user if a bindle is yanked, but the yanking host key is not known
+- MUST verify that the signature is valid (using the same formula described for regular signature validation).
+- SHOULD verify that the timestamp for yanking is not the same as the signature block's `at` timestamp.
+    - This is a trivial protection against attempts to forge a yank block.
+- SHOULD require that the host key used to yank is also the same host key originally used to sign
+    - This is complicated by host key rotation. Our recommendation is that on host key rotation, all packages should be re-signed using the new host key so that it is always clear who the host authority is.
```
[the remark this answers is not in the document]
technosophos: Yes, which is hypothetically what you would want to happen. In our case, host key rotation should be fairly rare. And you were right -- we might need a way to distribute host keys (but it can't be over the API because that is "in band" and hence not trustworthy).

docs/README.md
```diff
@@ -56,7 +56,7 @@ Here's a compact version of the above:
 ```console
 $ edit invoice.toml
 $ cat invoice.toml
-bindleVersion = "1.0.0"
+bindle_version = "1.0.0"
```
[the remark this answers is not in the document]
technosophos: Oh no... I bet it was in WAGI that we swiched from camel case to snake case.

docs/signing-spec.md
technosophos: Fixed

docs/signing-spec.md
technosophos: The text says they should sign as an `approver` and not alter the `creator` role. So the same _person_ would sign the same bindle two different times... but with two different keys (the old and the new) at two different times.

docs/signing-spec.md
technosophos: Okay, for now I have noted the problem, but I don't have a solution.

docs/README.md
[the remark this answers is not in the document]
technosophos: Okay, I reverted the change here. It doesn't feel like it belongs in this PR

docs/signing-spec.md
technosophos: I will rephrase... but basically, you could perform a sort of denial of service attack on a package by signing it with a known key, but with a broken signature.

So, for example, an attacker could get ahold of Alice's public key, and if Alice had the ability to sign a package, they could create a broken signature with Alice's public key and completely garbage signature data.

docs/signing-spec.md
technosophos: I think `creator` is a one-time thing, and I think we should honestly enforce that restriction when the API accepts the invoice.

The only case I can think of where an `approver` would want to modify is if the `approver` discovered a package was not to be trusted, but could not cause the `host` to yank the bindle.

I am less certain about proxies and hosts, where key rotation is desirable. During key rotation, I _think_ that we would want to remove the old key because it is no longer to be trusted. But maybe there would actually be no harm in appending a new signature and not removing the old one.

technosophos: Updated that signatures are append-only. This sounds like the right default option. We can relax iff necessary.

thomastaylor312 (approved):

PR #97, opened 2021-03-30
Fix versions in invoice spec sample

itowlson:
The version text in the sample in the invoice spec had gotten out of sync with modern thinking, and caused errors if you tried to serve the invoice from the spec.  This PR aligns the sample to the (presumably validated) test data.


technosophos (approved):
LGTM

Issue #87, opened 2021-02-03
Configuration Files and Config Dir


technosophos:
Bindle should have a configuration directory with the following files:

- server.toml: Basic configuration, like port, hostname, etc.
- keyring.toml: Public keyrings (possibly temporary)
- signing.key: mode 600 file with just the secret key used for signing packages

The configuration directory should default to a well-known location (either $XDG_CONFIG_DIR or /etc). IT should be automatically created if it does not already exist. And this directory name should be overrideable from CLI flags.

For starters, we don't need the keyring or key file -- just server.toml.

We may also want/need a client.toml for configuring a Bindle client.

technosophos:
- `server.toml` is done in #99
- `bindle create-key` does the logic for creating a signing key

So I think all that is left as of this writing is the public keyring

thomastaylor312:
After the latest PR merges with keychain management from the CLI, I think this has been addressed

PR #99, opened 2021-03-31
feat(server): Add optional config file

adamreese:
Ref: #87

Default config file name is `server.toml` in the directory set by the
`dirs` crate.

  https://docs.rs/dirs/3.0.1/dirs/fn.config_dir.html

### DIFF
--- a/bin/server.rs
+++ b/bin/server.rs
@@ -22,19 +22,17 @@ struct Opts {
         short = 'i',
         long = "address",
         env = "BINDLE_IP_ADDRESS_PORT",
-        default_value = "127.0.0.1:8080",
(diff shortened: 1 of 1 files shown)


bin/server.rs
```diff
@@ -90,3 +144,7 @@ async fn main() -> anyhow::Result<()> {
     )
     .await
 }
+
+fn default_config_file() -> Option<PathBuf> {
+    dirs::config_dir().map(|v| v.join("bindle/server.toml"))
```
technosophos, line 149: I :heart: the fact that config_dir() is built-in.

technosophos (approved):
LGTM

PR #100, opened 2021-03-31
added library support for yanked signature and keyring

technosophos:
This brings the current signature library into parity with the current signing spec.

- Adds yanked_signature
- Adds keyring
- Adds key entry
- Adds key entry signing
- Fixes a snake_case typo in the spec

### DIFF
--- a/src/invoice/signature.rs
+++ b/src/invoice/signature.rs
@@ -1,9 +1,13 @@
 //! Contains the Signature type along with associated types and Roles
 
+use ed25519_dalek::{Keypair, Signer};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 use std::fmt::{Display, Formatter, Result as FmtResult};
+
+/// The latest key ring version supported by this library.
+pub const KEY_RING_VERSION: &str = "1.0";
 
 /// A signature describes a cryptographic signature of the parcel list.
 ///
@@ -58,6 +62,7 @@ pub enum SignatureRole {
     Creator,
     Proxy,
     Host,
+    Approver,
 }
 
 impl Display for SignatureRole {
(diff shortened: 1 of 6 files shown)


src/invoice/signature.rs
```diff
@@ -69,7 +74,84 @@ impl Display for SignatureRole {
...
+///
+/// The purpose of this keyring is to validate signatures. The keyring NEVER
+/// contains private keys.
+///
+/// The concepts are described in the signing-spec.md document for Bindle.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct KeyRing {
+    pub version: String,
+    pub key: Vec<KeyEntry>,
```
radu-matei, line 93: nit: Should this be `keys`?

thomastaylor312: Yeah, probably `keys`

technosophos: So... no... because TOML.

```toml
[[key]]
label = "Brigade pipeline at builder.example.com"
roles = ["proxy"]
key = "bb453q4..."
labelSignature = "dsaff678..."
```

technosophos: This is the pattern used throughout Bindle: Use singular names for entries, and then mirror that name in the code. I suppose we could debate changing this (and I did originally start with Serde aliases, and revert). But the current PR is consistent with the rest of Bindle.

thomastaylor312 (approved):

src/invoice/signature.rs
radu-matei: We _could_ rename the field on the struct, if `keys` made sense in code.

radu-matei (approved):

src/invoice/signature.rs
thomastaylor312: Oh right, I remember that debate on parcels

PR #109, opened 2021-04-01
added a verify_label function to the key entry

technosophos:
This is an addition to the Keyring code, adding a way to verify that the label on a key entry has not been tampered with.

### DIFF
--- a/src/invoice/signature.rs
+++ b/src/invoice/signature.rs
@@ -134,10 +134,12 @@ impl KeyEntry {
     }
     pub fn verify_label(self, key: PublicKey) -> anyhow::Result<()> {
         match self.label_signature {
-            // If there is no label signature, then I guess we say everything's good.
-            None => Ok(()),
+            None => {
+                log::info!("Label was not signed. Skipping.");
+                Ok(())
+            }
             Some(txt) => {
-                let decoded_txt = base64::decode(txt.as_bytes())?;
+                let decoded_txt = base64::decode(txt)?;
                 let sig = EdSignature::new(decoded_txt.as_slice().try_into()?);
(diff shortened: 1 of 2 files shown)


src/invoice/signature.rs
```diff
@@ -131,6 +132,18 @@ impl KeyEntry {
         let sig = key.sign(self.label.as_bytes());
         self.label_signature = Some(base64::encode(sig.to_bytes()));
     }
+    pub fn verify_label(self, key: PublicKey) -> anyhow::Result<()> {
+        match self.label_signature {
+            // If there is no label signature, then I guess we say everything's good.
```
thomastaylor312, line 137: Should we `warn!`?

src/invoice/signature.rs
```diff
@@ -131,6 +132,18 @@ impl KeyEntry {
...
+            None => Ok(()),
+            Some(txt) => {
+                let decoded_txt = base64::decode(txt.as_bytes())?;
```
thomastaylor312, line 140: Nit: I don't think you need `as_bytes` here? `decode` takes `AsRef<[u8]>` which `str` implements

thomastaylor312 (approved):

src/invoice/signature.rs
technosophos: I'm not sure. Here's the thing: The signature is both optional and voluntary. It's basically for moving keyrings from one person to another and keeping trust. So a warning seems heavy-handed. When the file is on your localhost, I _think_ it is fine for the user to edit the label on their own. And they should be able to do that without then having to go generate a new crypto signature... since it's really just a local file at that point.

technosophos: For now, I'll `info!()` it, since that seems to split the difference.

PR #110, opened 2021-04-02
Add CLI command to create keyfile.

### LINKED ISSUE #103: Client needs to generate signing key
The client needs to generate a signing key for the `creator` role exactly once. We should probably add a command for this.

```
$ bindle create-key 'My Name <my@email>'
```

The key should be stored in the Bindle client's config directory with a name like `client-key`/`creator-key` or something to distinguish it from a server key. The mode of this file MUST be set to something like 600 or 700. Alternately, we could store it in the system keyring if that is broadly supported.

I think that the file could be as simple as a base64-encoded (or PEM-encoded if you prefer for some reason) bytes of the keypair. Possibly a better design would be to store it in TOML in a format similar to the `keyring.toml` described in the spec, but with a private key. This might allow a single client to have multiple keys (e.g. one for creator, one for approver, etc. or one for current, multiple for rotated).

Key generation code is all in `signature.rs`.

While we should support key rotation in the future, we probably don't need to do so now.

technosophos:
Add support for creating and reading keyfiles. Adds `bindle create-key` command.

Closes #103

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -1,7 +1,8 @@
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use bindle::client::{Client, ClientError, Result};
+use bindle::invoice::signature::{SecretKeyEntry, SecretKeyFile};
 use bindle::provider::ProviderError;
 use bindle::standalone::{StandaloneRead, StandaloneWrite};
 use bindle::{
@@ -107,6 +108,50 @@ async fn main() -> std::result::Result<(), ClientError> {
             )
             .await?;
             println!("{}", toml::to_string_pretty(&label)?);
+        }
+        SubCommand::CreateKey(create_opts) => {
+            let dir = create_opts.secret_file.unwrap_or_else(|| {
+                default_config_dir()
+                    .unwrap_or_else(|| PathBuf::from("."))
+                    .join("secret_keys.toml")
+            });
+            println!("Writing keys to {}", dir.to_string_lossy());
+
+            match std::fs::metadata(dir.clone()) {
+                Err(_) => {
+                    println!(
+                        "File {} does not exist. Creating it.",
+                        dir.to_string_lossy()
+                    );
+                    let mut keyfile = SecretKeyFile::default();
+                    let newkey = SecretKeyEntry::new(
+                        create_opts.label,
+                        vec![bindle::SignatureRole::Creator],
+                    );
+                    keyfile.key.push(newkey);
+                    keyfile
+                        .save_file(dir)
+                        .map_err(|e| ClientError::Other(e.to_string()))?;
+                }
+                Ok(info) => {
+                    if !info.is_file() {
+                        eprint!("Path must point to a file.");
+                        return Err(ClientError::Other(
+                            "Keyfile cannot be directory or symlink".to_owned(),
+                        ));
+                    }
+                    let mut keyfile = SecretKeyFile::load_file(dir.clone())
+                        .map_err(|e| ClientError::Other(e.to_string()))?;
+                    let newkey = SecretKeyEntry::new(
+                        create_opts.label,
+                        vec![bindle::SignatureRole::Creator],
+                    );
+                    keyfile.key.push(newkey);
+                    keyfile
+                        .save_file(dir)
+                        .map_err(|e| ClientError::Other(e.to_string()))?;
+                }
+            }
         }
     }
 
@@ -263,4 +308,8 @@ fn map_storage_error(e: ProviderError) -> ClientError {
         ProviderError::ProxyError(inner) => inner,
         _ => ClientError::Other(format!("{:?}", e)),
     }
+}
+
+fn default_config_dir() -> Option<PathBuf> {
+    dirs::config_dir().map(|v| v.join("bindle"))
 }
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -151,7 +151,7 @@ async fn main() -> std::result::Result<(), ClientError> {
                         .await
                         .map_err(|e| ClientError::Other(e.to_string()))?;
                 }
-                Err(e) => return Err(ClientError::Other(e.to_string())),
+                Err(e) => return Err(e.into()),
             }
         }
     }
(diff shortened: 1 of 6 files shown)


bin/client/main.rs
```diff
@@ -108,6 +109,50 @@ async fn main() -> std::result::Result<(), ClientError> {
             .await?;
             println!("{}", toml::to_string_pretty(&label)?);
         }
+        SubCommand::CreateKey(create_opts) => {
+            let dir = create_opts.secret_file.unwrap_or_else(|| {
+                default_config_dir()
+                    .unwrap_or_else(|| PathBuf::from("."))
+                    .join("secret_keys.toml")
+            });
+            println!("Writing keys to {}", dir.to_string_lossy());
```
thomastaylor312, line 118: I think `dir.display()` is preferred here rather than doing a string conversion. Same goes on all other similar lines

bin/client/main.rs
```diff
@@ -108,6 +109,50 @@ async fn main() -> std::result::Result<(), ClientError> {
...
+
+            match std::fs::metadata(dir.clone()) {
```
thomastaylor312, line 120: This should be `tokio::fs::metadata` as we are in an async function

thomastaylor312: Nit: Oh also, I think you can just borrow `dir` here rather than cloning

bin/client/main.rs
```diff
@@ -108,6 +109,50 @@ async fn main() -> std::result::Result<(), ClientError> {
...
+                Err(_) => {
```
thomastaylor312, line 121: This should have a guard clause that checks if the Error is a not found error as other errors we probably want to bail out on: `Err(e) if matches!(e.kind(), ErrorKind::NotFound)`

bin/client/main.rs
```diff
@@ -108,6 +109,50 @@ async fn main() -> std::result::Result<(), ClientError> {
...
+                    keyfile.key.push(newkey);
+                    keyfile
+                        .save_file(dir)
+                        .map_err(|e| ClientError::Other(e.to_string()))?;
+                }
+                Ok(info) => {
+                    if !info.is_file() {
+                        eprint!("Path must point to a file.");
+                        return Err(ClientError::Other(
+                            "Keyfile cannot be directory or symlink".to_owned(),
```
thomastaylor312, line 140: Nit: I think we might want to support symlinks if they point at a regular file

src/invoice/signature.rs
```diff
@@ -133,6 +134,64 @@ impl KeyEntry {
...
+    fn default() -> Self {
+        Self {
+            version: KEY_RING_VERSION.to_owned(),
+            key: vec![],
+        }
+    }
+}
+
+impl SecretKeyFile {
+    pub fn load_file(path: PathBuf) -> Result<SecretKeyFile, anyhow::Error> {
```
thomastaylor312, line 181: This should be async and use the tokio::fs functions

thomastaylor312: Nit: Use `anyhow::Result` instead to match usage below

src/invoice/signature.rs
```diff
@@ -133,6 +134,64 @@ impl KeyEntry {
...
+        let s = std::fs::read_to_string(path)?;
+        let t = toml::from_str(s.as_str())?;
+        Ok(t)
+    }
+
+    /// Save the present keyfile to the named path.
+    pub fn save_file(&self, dest: PathBuf) -> anyhow::Result<()> {
```
thomastaylor312, line 188: Should also be async like above

bin/client/opts.rs
```diff
@@ -66,6 +66,11 @@ pub enum SubCommand {
         about = "generates a label for the given file and prints it to stdout. This can be used to generate the label and add it to an invoice"
     )]
     GenerateLabel(GenerateLabel),
+    #[clap(
+        name = "create-key",
+        about = "creates a new signing key and places it in the local secret keys. If no secret file is provided, this will store in the default config directory for Bindle. The LABEL is typically a name and email addres, of the form 'name <email>'."
```
thomastaylor312, line 71: s/addres/address

bin/client/opts.rs
```diff
@@ -225,6 +230,22 @@ pub struct GenerateLabel {
...
+    #[clap(
+        index = 1,
+        value_name = "LABEL",
+        about = "The name of the key, such as 'Matt <me@example.com>'"
+    )]
+    pub label: String,
+    #[clap(
+        short = 'f',
+        long = "secrets-file",
+        about = "the path to the file where secrets should be stored. If it does not exist, it will be created. If it does exist, the key will be appened."
```
thomastaylor312, line 244: s/appened/appended

thomastaylor312 (changes requested):

bin/client/main.rs
technosophos: Storing a key file behind a symlink is insecure because of the ambiguities of both permissions and the underlying filesystem. (For example, a malicious user who can modify mount tables could trick the client into loading a different key or otherwise modify the file at the other end of the symlink). The recommendation is to never reference private keys via symlink.

bin/client/main.rs
```diff
@@ -108,6 +109,51 @@ async fn main() -> std::result::Result<(), ClientError> {
...
+                        ));
+                    }
+                    let mut keyfile = SecretKeyFile::load_file(dir.clone())
+                        .await
+                        .map_err(|e| ClientError::Other(e.to_string()))?;
```
thomastaylor312, line 143: I missed this before, so I am not going to block, but there is a `ClientError:Io` type. So you should be able to just do `e.into()`. I can't remember how smart the `?` operator is, but it might even be able to convert without the mapping

thomastaylor312 (approved):

bin/client/main.rs
technosophos: Since load_file() can cause IO, Dalek, and TOML/serde errors, it looks like I have to leave this as-is. But I did find another place where I can convert to a ClientError::Io.

thomastaylor312: We might want a new class of error eventually for Signature stuff. But obviously not right now

PR #117, opened 2021-04-09
add sign-invoice command

technosophos:
This is the third step toward implementing signing and verification. It uses a key created with `create-key` and signs an invoice.

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -2,7 +2,8 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use bindle::client::{Client, ClientError, Result};
-use bindle::invoice::signature::{SecretKeyEntry, SecretKeyFile};
+use bindle::invoice::signature::{Keypair, SecretKeyEntry, SecretKeyFile, SignatureRole};
+use bindle::invoice::Invoice;
 use bindle::provider::ProviderError;
 use bindle::standalone::{StandaloneRead, StandaloneWrite};
 use bindle::{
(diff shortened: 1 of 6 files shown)


bin/client/main.rs
```diff
@@ -91,6 +92,37 @@ async fn main() -> std::result::Result<(), ClientError> {
...
+            // Role
+            let role = if let Some(r) = sign_opts.role {
+                role_from_name(r)?
+            } else {
+                SignatureRole::Creator
+            };
+            // Keyfile
+            let keyfile = sign_opts.secret_file.unwrap_or_else(|| {
+                default_config_dir()
+                    .unwrap_or_else(|| PathBuf::from("."))
```
thomastaylor312, line 105: Nit: I think you can just do `PathBuf::from("secret_keys.toml")` here

src/invoice/mod.rs
```diff
@@ -130,8 +130,19 @@ impl Invoice {
         &mut self,
         signer_name: String,
         signer_role: SignatureRole,
-        key: Keypair,
+        key: &Keypair,
     ) -> Result<(), SignatureError> {
+        // The spec says it is illegal for the a single key to sign the same invoice
+        // more than once.
+        let encoded_key = base64::encode(key.public.to_bytes());
+        if let Some(sigs) = self.signature.as_ref() {
```
thomastaylor312, line 138: This `if let` and `for` + `if` block could be reduced to something close to this:

```rust
if self.signature.as_ref().map(|sigs|  sigs.iter().any(|s| s.key == encoded_key) ).unwrap_or(false) {
    return Err(SignatureError::DuplicateSignature)
}
```

thomastaylor312 (approved):
Some minor nits. Anything else I figured we'd fix/refactor in our pairing session

bin/client/main.rs
technosophos: I don't think so... that would give me `secret_keys.toml/secret_keys.toml`. The idea is to get the default config dir or else use `.`, and _then_ append `secret_keys.toml`. So if config dir is `/foo/bar`, the result should be `/foo/bar/secret_keys.toml`.

Or maybe I'm misunderstanding something about what you are suggesting?

thomastaylor312: Oh right, you are joining on the result of the operation. Ignore me

src/invoice/mod.rs
technosophos: This part is already re-written for the next PR... so gonna leave this for now and just merge it in the next PR

PR #120, opened 2021-04-14
ensure that a config directory exists before trying to use it

### LINKED ISSUE #116: Should ensure that `config_dir() + "bindle"` exists
Right now, Bindle doesn't check to see if the dir exists before attempting to operate on it. While this is fine for reads, for writes it is problematic. See `create-key` for an example.

technosophos:
Make sure that the config dir exists before attempting to use it.

Closes #116

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -100,11 +100,11 @@ async fn main() -> std::result::Result<(), ClientError> {
                 SignatureRole::Creator
             };
             // Keyfile
-            let keyfile = sign_opts.secret_file.unwrap_or_else(|| {
-                default_config_dir()
-                    .unwrap_or_else(|| PathBuf::from("."))
-                    .join("secret_keys.toml")
-            });
+            let keyfile = match sign_opts.secret_file {
+                Some(dir) => dir,
+                None => ensure_config_dir().await?.join("secret_keys.toml"),
+            };
+
             // Signing key
             let (key, label) = first_matching_key(keyfile, &role).await?;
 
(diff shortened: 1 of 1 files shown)


bin/client/main.rs
```diff
@@ -100,11 +100,11 @@ async fn main() -> std::result::Result<(), ClientError> {
                 SignatureRole::Creator
             };
             // Keyfile
-            let keyfile = sign_opts.secret_file.unwrap_or_else(|| {
-                default_config_dir()
-                    .unwrap_or_else(|| PathBuf::from("."))
-                    .join("secret_keys.toml")
-            });
+            let keyfile = match sign_opts.secret_file {
+                Some(dir) => dir,
```
thomastaylor312, line 104: Should you also ensure that this dir is created?

bin/client/main.rs
```diff
@@ -142,11 +142,10 @@ async fn main() -> std::result::Result<(), ClientError> {
             println!("{}", toml::to_string_pretty(&label)?);
         }
         SubCommand::CreateKey(create_opts) => {
-            let dir = create_opts.secret_file.unwrap_or_else(|| {
-                default_config_dir()
-                    .unwrap_or_else(|| PathBuf::from("."))
-                    .join("secret_keys.toml")
-            });
+            let dir = match create_opts.secret_file {
+                Some(dir) => dir,
```
thomastaylor312, line 146: Same question here

thomastaylor312 (commented):
One concern to discuss around ensuring the dir passed in by the user

bin/client/main.rs
technosophos: Same answer. ;-)

bin/client/main.rs
technosophos: I don't think so. My take is that if the user supplies a path, it is up to them to create the path. This is the way most programs work: If they use an `$XDG_*` dir, the program takes responsibility for creating (same for a hidden dir in `$HOME`). But if the user supplies a path, the tool assumes the user did the work.

In this particular case, per the Rust recommendations, we don't use `metadata()` and instead let the call to create the file be the point of failure.

thomastaylor312 (approved):

PR #127, opened 2021-04-15
Verify creator on invoice upload

technosophos:
This PR wires up all of the signing and verification logic for the server.

Todo:

- [x] Unit test for VerificationStrategy
- [ ] ~~Implementation for Proxy: Sign before sending upstream.~~ Moved to #141
- [x] Implementation for LRU
- [x] Remove `Invoice::verify()`, since that is now on VerificationStrategy
- [x] API server needs to be able to load keyfiles
- [ ] ~~API server needs to be able to load strategy~~ Moved to #143
- [x] File provider needs to load public keyring
- [x] Attempt to remove Dalek from public API

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -345,8 +345,7 @@ async fn load_keyring(keyring: Option<PathBuf>) -> anyhow::Result<KeyRing> {
     // This takes an Option<PathBuf> because we want to wrap all of the flag handling in this
     // function, including setting the default if the kyering is None.
     let dir = keyring
-        .clone()
-        .unwrap_or_else(|| default_config_dir().unwrap_or_else(|| PathBuf::from(".bindle")))
+        .unwrap_or_else(|| default_config_dir())
         .join("keyring.toml");
     let kr = bindle::client::load::toml(dir).await?;
     Ok(kr)
@@ -360,8 +359,10 @@ fn map_storage_error(e: ProviderError) -> ClientError {
     }
 }
 
-fn default_config_dir() -> Option<PathBuf> {
-    dirs::config_dir().map(|v| v.join("bindle"))
+fn default_config_dir() -> PathBuf {
+    dirs::config_dir()
+        .map(|v| v.join("bindle/"))
+        .unwrap_or_else(|| "./bindle".into())
 }
 
 /// Get the config dir, ensuring that it exists.
@@ -374,7 +375,7 @@ fn default_config_dir() -> Option<PathBuf> {
 ///
 /// This will return an error
 async fn ensure_config_dir() -> Result<PathBuf> {
-    let dir = default_config_dir().unwrap_or_else(|| PathBuf::from("./bindle"));
+    let dir = default_config_dir();
     tokio::fs::create_dir_all(dir.clone()).await?;
     Ok(dir)
 }
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -2,7 +2,9 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use bindle::client::{Client, ClientError, Result};
-use bindle::invoice::signature::{KeyRing, SecretKeyEntry, SecretKeyFile, SignatureRole};
+use bindle::invoice::signature::{
+    KeyRing, SecretKeyEntry, SecretKeyFile, SecretKeyStorage, SignatureRole,
+};
 use bindle::invoice::Invoice;
 use bindle::provider::ProviderError;
 use bindle::standalone::{StandaloneRead, StandaloneWrite};
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -2,7 +2,7 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use bindle::client::{Client, ClientError, Result};
-use bindle::invoice::signature::{Keypair, SecretKeyEntry, SecretKeyFile, SignatureRole};
+use bindle::invoice::signature::{KeyRing, SecretKeyEntry, SecretKeyFile, SignatureRole};
 use bindle::invoice::Invoice;
 use bindle::provider::ProviderError;
 use bindle::standalone::{StandaloneRead, StandaloneWrite};
@@ -39,6 +39,9 @@ async fn main() -> std::result::Result<(), ClientError> {
     let local = bindle::provider::file::FileProvider::new(
         bindle_dir,
         bindle::search::NoopEngine::default(),
+        load_keyring(opts.keyring)
+            .await
+            .unwrap_or_else(|_| KeyRing::default()),
     )
     .await;
     let proxy = bindle::proxy::Proxy::new(bindle_client.clone());
@@ -334,6 +337,17 @@ async fn get_all<C: Cache + Send + Sync + Clone>(cache: C, opts: Get) -> Result<
     }
 
     Ok(())
+}
+
+async fn load_keyring(keyring: Option<PathBuf>) -> anyhow::Result<KeyRing> {
+    // This takes an Option<PathBuf> because we want to wrap all of the flag handling in this
+    // function, including setting the default if the kyering is None.
+    let dir = keyring
+        .clone()
+        .unwrap_or_else(|| default_config_dir().unwrap_or_else(|| PathBuf::from(".bindle")))
+        .join("keyring.toml");
+    let kr = bindle::client::load::toml(dir).await?;
+    Ok(kr)
 }
 
 fn map_storage_error(e: ProviderError) -> ClientError {
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -108,11 +108,11 @@ async fn main() -> std::result::Result<(), ClientError> {
             };
 
             // Signing key
-            let (key, label) = first_matching_key(keyfile, &role).await?;
+            let key = first_matching_key(keyfile, &role).await?;
 
             // Load the invoice and sign it.
             let mut inv: Invoice = bindle::client::load::toml(sign_opts.invoice.as_str()).await?;
-            inv.sign(label, role.clone(), &key)?;
+            inv.sign(key.label.clone(), role.clone(), &key)?;
 
             // Write the signed invoice to a file.
             let outfile = sign_opts
@@ -373,7 +373,7 @@ fn role_from_name(name: String) -> Result<SignatureRole> {
     }
 }
 
-async fn first_matching_key(fpath: PathBuf, role: &SignatureRole) -> Result<(Keypair, String)> {
+async fn first_matching_key(fpath: PathBuf, role: &SignatureRole) -> Result<SecretKeyEntry> {
     let keys = SecretKeyFile::load_file(fpath.clone()).await.map_err(|e| {
         ClientError::Other(format!(
             "Error loading file {}: {}",
@@ -381,11 +381,8 @@ async fn first_matching_key(fpath: PathBuf, role: &SignatureRole) -> Result<(Key
             e.to_string()
         ))
     })?;
-    for key in keys.key {
-        if key.roles.contains(role) {
-            let pair = key.key().map_err(|e| ClientError::Other(e.to_string()))?;
-            return Ok((pair, key.label));
-        }
-    }
-    Err(ClientError::Other("No satisfactory key found".to_owned()))
+
+    keys.get_first_matching(role)
+        .map(|k| k.to_owned())
+        .ok_or_else(|| ClientError::Other("No satisfactory key found".to_owned()))
 }
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -112,7 +112,7 @@ async fn main() -> std::result::Result<(), ClientError> {
 
             // Load the invoice and sign it.
             let mut inv: Invoice = bindle::client::load::toml(sign_opts.invoice.as_str()).await?;
-            inv.sign(key.label.clone(), role.clone(), &key)?;
+            inv.sign(role.clone(), &key)?;
 
             // Write the signed invoice to a file.
             let outfile = sign_opts
--- a/bin/client/opts.rs
+++ b/bin/client/opts.rs
@@ -25,6 +25,14 @@ pub struct Opts {
         about = "The directory where bindles are stored/cached, defaults to $XDG_CACHE_HOME"
     )]
     pub bindle_dir: Option<PathBuf>,
+
+    #[clap(
+        short = 'r',
+        long = "keyring",
+        about = "The path to the keyring file. Defaults to $XDG_CONFIG/bindle/keyring.toml"
+    )]
+    pub keyring: Option<PathBuf>,
+
     #[clap(subcommand)]
     pub subcmd: SubCommand,
 }
--- a/bin/server.rs
+++ b/bin/server.rs
@@ -7,6 +7,7 @@ use bindle::{
     invoice::signature::KeyRing,
     provider, search,
     server::{server, TlsConfig},
+    signature::SecretKeyFile,
 };
 
 const DESCRIPTION: &str = r#"
@@ -65,6 +66,14 @@ struct Opts {
         about = "the path to the keyring file"
     )]
     keyring_file: Option<PathBuf>,
+
+    #[clap(
+        name = "signing_keys",
+        long = "signing-keys",
+        env = "BINDLE_SIGNING_KEYS",
+        about = "location of the TOML file that holds the signing keys"
+    )]
+    signing_file: Option<PathBuf>,
 }
 
 #[tokio::main]
@@ -157,6 +166,19 @@ async fn main() -> anyhow::Result<()> {
         Err(e) => anyhow::bail!("failed to read file {}: {}", keyring_file.display(), e),
     };
 
+    let signing_keys: PathBuf = opts
+        .signing_file
+        .or_else(|| {
+            config
+                .get("signing-keys")
+                .map(|v| v.to_string().parse().unwrap())
+        })
+        .unwrap_or(PathBuf::from(
+            default_config_dir()
+                .unwrap_or_else(|| PathBuf::from("./bindle"))
+                .join("signing-keys.toml"),
+        ));
+
     let cert_path = opts.cert_path.or_else(|| {
         config
             .get("cert-path")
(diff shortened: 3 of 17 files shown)


technosophos:
Breaking out proxy task to a separate issue: #141 

Removing LRU because I don't think there is any applicable role for signing a cached record. But #141 might change that.

technosophos:
I moved the loadable verification strategy to #143 because this PR doesn't yet have verification in it. It adds support for host signing, but not most forms of verification.

src/invoice/signature.rs
```diff
@@ -224,6 +346,31 @@ impl SecretKeyFile {
...
 }
 
+impl SecretKeyStorage for SecretKeyFile {
+    fn get_first_matching(&self, role: &SignatureRole) -> Option<&SecretKeyEntry> {
+        self.key.iter().find(|k| k.roles.contains(role)).clone()
+    }
+}
+
+pub struct SecretKeyFileLoader {
+    path: PathBuf,
```
thomastaylor312, line 356: This is a private data member but I don't see a `new` or other type of constructor here for someone using it publicly

src/invoice/verification.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// AuthoritativeIntegrity verifies that at least one of the Creator or Approver keys
+    /// is known and the signature is valid.
+    AuthoritativeIntegrity,
+    /// Verify that the Creator key is known and that all signatures are valid.
+    ///
+    /// This is subject to a DOS attack if a signer can generate intentionally bad signatures.
+    GreedyVerification,
+    /// Verify that every key on the invoice is known, and that every signature is valid.
+    ExhaustiveVerification,
+    /// Verifies that all signatures of the given roles are valid and signed by know keys.
```
thomastaylor312, line 23: Spelling: known

src/invoice/verification.rs
```diff
@@ -0,0 +1,389 @@
...
+    ///
+    /// If the bool is true, unknown signers will also be valdidated.
+    /// Be aware that doing so may make the validation subject to a special for of
```
thomastaylor312, line 26: Spelling: form

src/invoice/verification.rs
```diff
@@ -0,0 +1,389 @@
...
+                    SignatureRole::Creator,
+                    SignatureRole::Approver,
+                    SignatureRole::Host,
+                    SignatureRole::Proxy,
+                ],
+                true,
+                true,
+                false,
+            ),
+            VerificationStrategy::MultipleAttestation(a, b) => (a.clone(), b.clone(), true, true),
```
thomastaylor312, line 99: Nit: I don't think you need `b.clone()` because `bool` implements `Copy`

src/invoice/verification.rs
```diff
@@ -0,0 +1,389 @@
...
+        };
+
+        // Either the Creator or an Approver must be in the keyring
+        match inv.signature.as_ref() {
+            None => Ok(()),
```
thomastaylor312, line 104: Should we at least log a warning if there are no signatures? Or is this going to go away once we fully enforce things?

src/invoice/verification.rs
```diff
@@ -0,0 +1,389 @@
...
+    ///
+    /// If no signatures are on the invoice, this will succeed.
+    ///
+    /// A strategy will determine success or failure based on whether the signature is verified,
+    /// whether the keys are known, whether the requisite number/roles are satisfied, and
+    /// so on.
+    pub fn verify(&self, inv: &Invoice, keyring: &KeyRing) -> Result<(), SignatureError> {
+        let (roles, all_valid, all_verified, all_roles) = match self {
+            VerificationStrategy::GreedyVerification => {
+                (vec![SignatureRole::Creator], true, true, true)
```
thomastaylor312, line 77: Nit: Based on what I am seeing in the code below, I don't think we need to reallocate owned `Vec`s every time we verify. Not sure if we can do this with enum variants, but it might be worth looking into making each of these sets a `const` slice and then using those in the logic

src/invoice/verification.rs
```diff
@@ -0,0 +1,389 @@
...
+        };
+
+        // Either the Creator or an Approver must be in the keyring
+        match inv.signature.as_ref() {
+            None => Ok(()),
+            Some(signatures) => {
+                let mut known_key = false;
+                let mut filled_roles: Vec<SignatureRole> = vec![];
+                for s in signatures {
+                    debug!(by = %s.by, "Checking signature");
```
thomastaylor312, line 109: Super nit: Consider setting up a span for each loop so the `by` value can be set on each logged field

src/invoice/verification.rs
```diff
@@ -0,0 +1,389 @@
...
+                    if !target_role && !all_verified {
+                        debug!("Not a target role, not checking for verification");
+                        continue;
+                    } else if all_roles {
+                        filled_roles.push(role);
+                    }
+                    // See if the public key is known to us
+                    let pubkey = base64::decode(s.key.clone())
+                        .map_err(|_| SignatureError::CorruptKey(s.key.to_string()))?;
+                    let pko = PublicKey::from_bytes(pubkey.as_slice())
```
thomastaylor312, line 139: `pko` == "Public key owned"?

src/invoice/verification.rs
```diff
@@ -0,0 +1,389 @@
...
+    use super::*;
+    use crate::invoice::*;
+
+    #[test]
+    fn test_verification_strategies() {
+        let invoice = r#"
+        bindleVersion = "1.0.0"
+
+        [bindle]
+        name = "aricebo"
```
thomastaylor312, line 191: "arecibo"

src/provider/file/mod.rs
```diff
@@ -151,13 +157,24 @@ impl<T: Search + Send + Sync> FileProvider<T> {
 #[async_trait::async_trait]
 impl<T: crate::search::Search + Send + Sync> Provider for FileProvider<T> {
     #[instrument(level = "trace", skip(self, inv), fields(id = %inv.bindle.id))]
-    async fn create_invoice(&self, inv: &crate::Invoice) -> Result<Vec<crate::Label>> {
+    async fn create_invoice(
+        &self,
+        inv: &mut crate::Invoice,
+        role: SignatureRole,
+        secret_key: &SecretKeyEntry,
+        strategy: VerificationStrategy,
```
thomastaylor312, line 165: Nit: Does it make sense for `create_invoice` to own `strategy` or can we just do a borrow? Looks like we could just do a borrow, but maybe that doesn't matter too much

src/provider/file/mod.rs
```diff
@@ -599,19 +622,41 @@ mod test {
...
         )
         .await;
         let inv_name = scaffold.invoice.canonical_name();
+
+        let sk = mock_secret_key();
         // Create an file
-        let missing = store.create_invoice(&scaffold.invoice).await.unwrap();
+        let missing = store
+            .create_invoice(
+                &mut scaffold.invoice,
```
thomastaylor312, line 653: Ok, so this made me realize that this pattern of `&mut invoice` is really odd. Basically we end up mutating the invoice the user gives us rather than having them depend on the returned one for getting the signed invoice. Although I don't think it has to happen here, we should probably change this to be `mut invoice:Invoice` and just have it take ownership. I can do that in a general "cloning cleanup" I'd like to do over the whole repo anyway once this is merged

src/provider/mod.rs
```diff
@@ -51,7 +52,36 @@ pub trait Provider {
...
+        &self,
+        inv: &mut super::Invoice,
+        role: SignatureRole,
+        secret_key: &SecretKeyEntry,
+    ) -> Result<()> {
+        inv.sign(role, secret_key)?;
+        Ok(())
+    }
+
+    // TODO: This can probably be removed.
```
thomastaylor312, line 75: Any clarification on this comment? Are providers supposed to require verification when implementing?

src/proxy/mod.rs
```diff
@@ -23,7 +27,20 @@ impl Proxy {
...
 impl Provider for Proxy {
-    async fn create_invoice(&self, inv: &crate::Invoice) -> Result<Vec<crate::Label>> {
+    async fn create_invoice(
+        &self,
+        inv: &mut crate::Invoice,
+        _role: SignatureRole,
+        secret_key: &SecretKeyEntry,
+        strategy: VerificationStrategy,
+    ) -> Result<Vec<crate::Label>> {
+        // TODO: Need a way to load the local public keyring
```
thomastaylor312, line 37: Maybe change the `new` function? Does this need to be addressed before merge?

src/server/reply.rs
```diff
@@ -120,6 +120,7 @@ pub fn into_reply(error: ProviderError) -> warp::reply::WithStatus<SerializedDat
         #[cfg(feature = "client")]
         ProviderError::ProxyError(_) => StatusCode::INTERNAL_SERVER_ERROR,
         ProviderError::Other(_) | ProviderError::Io(_) => StatusCode::INTERNAL_SERVER_ERROR,
+        ProviderError::FailedSigning(_) => StatusCode::BAD_REQUEST,
```
thomastaylor312, line 123: Is there ever a time where a failed signing could be caused by server misconfiguration? Could be good to differentiate in that case

thomastaylor312 (changes requested):
Most of these are nits or minor changes. A few small fixes required. Otherwise this is looking really good from a code perspective

bin/client/main.rs
```diff
@@ -334,6 +339,17 @@ async fn get_all<C: Cache + Send + Sync + Clone>(cache: C, opts: Get) -> Result<
     Ok(())
 }
 
+async fn load_keyring(keyring: Option<PathBuf>) -> anyhow::Result<KeyRing> {
+    // This takes an Option<PathBuf> because we want to wrap all of the flag handling in this
+    // function, including setting the default if the kyering is None.
+    let dir = keyring
+        .clone()
```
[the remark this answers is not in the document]
technosophos: Fixed

src/invoice/mod.rs
```diff
@@ -325,47 +260,52 @@ mod test {
...
 
         println!("{}", toml::to_string(&invoice).unwrap());
 
         // There should be two signature blocks
         assert_eq!(2, invoice.signature.as_ref().unwrap().len());
 
         // With the keyring, the signature should work
-        invoice
-            .verify(keyring)
+        println!("==== IN");
```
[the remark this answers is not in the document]
technosophos: Removed

src/invoice/signature.rs
```diff
@@ -106,6 +109,32 @@ impl Default for KeyRing {
...
+        }
+    }
+    pub fn contains(&self, key: &PublicKey) -> bool {
+        // This could definitely be optimized.
+        for k in self.key.clone() {
+            // Note that we are skipping malformed keys because they don't matter
+            // when doing a contains(). If they key is malformed, it definitely
+            // is not the key we are looking for.
+            match k.public_key() {
+                Err(e) => println!("Error parsing key: {}", e),
```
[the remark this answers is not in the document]
technosophos: Fixed

src/invoice/signature.rs
```diff
@@ -193,6 +281,40 @@ impl SecretKeyEntry {
...
+    /// In general, if multiple keys match, the implementation chooses the "best fit"
+    /// and returns that key.
+    fn get_first_matching(&self, role: &SignatureRole) -> Option<&SecretKeyEntry>;
+}
+
+/// This is a speciality wrapper for loading secrets files at the last moment each time a key is accessed.
+///
+/// Note that this logs and swallows errors, but is more secure than loading keys once.
+// #[derive(Debug, Clone)]
+// pub struct LazySecretKeyLoader {
```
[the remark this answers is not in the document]
technosophos: Fixed

src/invoice/signature.rs
technosophos: Removed

src/invoice/verification.rs
technosophos: Fixed

src/invoice/verification.rs
technosophos: I will log an info for now. But, yeah, eventually the server will not accept invoices that do not have signatures. This PR just does the host signature. A different one will have to do the creator signature.

src/invoice/verification.rs
technosophos: Alright, I tried and it didn't seem to work the way I thought. So for now I am gonna skip this.

src/invoice/verification.rs
technosophos: Public Key Object

src/invoice/verification.rs
technosophos: fixed

src/provider/mod.rs
technosophos: Removed.

src/proxy/mod.rs
technosophos: That was originally for #141... but I might just remove it now and re-do it later. Right now it basically does nothing.

technosophos: Actually... I am going to leave this for pairing. I am not sure I completely understand the Proxy setup

src/server/reply.rs
technosophos: The only way to trigger this failure is if a signature fails to match. That could be because the signature is corrupt or because the key is corrupt. But we don't want to tell the user which because that is an "information disclosure" that could be used in an exploit.

bin/server.rs
```diff
@@ -106,6 +124,59 @@ async fn main() -> anyhow::Result<()> {
...
+    let keyring_file: PathBuf = opts
+        .keyring_file
+        .or_else(|| {
+            config
+                .get("keyring")
+                .map(|v| v.as_str().unwrap().parse().unwrap())
+        })
+        .unwrap_or(PathBuf::from(
+            default_config_dir()
+                .unwrap_or_else(|| PathBuf::from("./bindle"))
```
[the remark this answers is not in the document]
technosophos: I find it frustrating that `default_config_dir()` returns an `Option`. Shouldn't a default function return an actual thing? It seems super silly to have to keep saying, in essence, "give me the default directory or here's a default default"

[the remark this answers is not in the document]
technosophos: I'm just gonna make that change. It makes more sense

bin/client/opts.rs
```diff
@@ -25,6 +25,14 @@ pub struct Opts {
         about = "The directory where bindles are stored/cached, defaults to $HOME/.bindle/bindles"
     )]
     pub bindle_dir: Option<PathBuf>,
+
+    #[clap(
+        short = 'r',
+        long = "keyring",
+        about = "The path to the keyring file. Defaults to $XDG_CONFIG/bindle/keyring.toml"
```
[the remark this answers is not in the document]
technosophos: It should be XDG_CONFIG/bindle/keyring.toml. There were a few cases where `.bindle` was used because we used to use it (unwrap_or_default() will not join its value... so what you were reading was "XDG_CONFIG/bindle OR .bindle".

bin/server.rs
```diff
@@ -143,10 +223,62 @@ async fn main() -> anyhow::Result<()> {
...
+
+async fn ensure_config_dir() -> anyhow::Result<PathBuf> {
+    let dir = default_config_dir().unwrap_or_else(|| PathBuf::from("./bindle"));
+    tokio::fs::create_dir_all(dir.clone()).await?;
+    Ok(dir)
+}
+
+async fn ensure_signing_keys() -> anyhow::Result<PathBuf> {
+    let base = ensure_config_dir().await?;
+    let signing_keyfile = base.join("signing-keys.toml");
```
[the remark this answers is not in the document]
technosophos: I looked through the code and don't see a lot of cases where these are repeated ("keyring.toml" appears in server.rs and main.rs, but those are different crates). So I don't think it makes sense to bump them up to constants.

src/invoice/signature.rs
```diff
@@ -149,6 +192,51 @@ impl KeyEntry {
             }
         }
     }
+    pub(crate) fn public_key(&self) -> Result<PublicKey, SignatureError> {
```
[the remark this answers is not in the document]
technosophos: No, not really. Most of the time, the file is deserialized and only one or two keys are ever actually used. So deserializing them all is probably more work than it's worth. Further, if a file contains a bad public key (which happens), it doesn't prevent the entire keyring from loading until someone tries to use the key. Given that this file can be manually edited, that is probably a good thing.

src/provider/file/mod.rs
technosophos: I'm not sure why I would prefer a borrow in this case. Can you explain why that would be better?

src/provider/file/mod.rs
technosophos: Okay. I will leave it to you to change it however you want it.

src/proxy/mod.rs
technosophos: Alright, I figured out what I was doing. It is part of #141 that I need to redo this. So I have removed it for now and left a TODO to do it along with #141.

src/provider/file/mod.rs
thomastaylor312: I think of it from a consumer point of view. If I have a verification strategy I created, now I have to clone it if I plan on using it anywhere else when I call this function. As far as I could tell, this function doesn't have any need for ownership (it isn't storing it somewhere, and is just using it to access data). With something like this very lightweight enum, it doesn't _truly_ matter, hence the nit. But in general I like to think of the question "Do I need to take ownership from the user?"

bin/server.rs
```diff
@@ -111,6 +129,55 @@ async fn main() -> anyhow::Result<()> {
...
+    //
+    // All other cases are considered errors worthy of failing.
+    let keyring: KeyRing = match std::fs::metadata(&keyring_file) {
+        Ok(md) if md.is_file() => load_toml(keyring_file).await?,
+        Ok(_) => Err(anyhow::anyhow!(
+            "Expected {} to be a regular file",
+            keyring_file.display()
+        ))?,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            println!("No keyring.toml found. Using default keyring.");
```
thomastaylor312, line 160: This should probably be a `warn!`

thomastaylor312 (approved):
Whew, this was a beast. Thanks for slogging through my comments

bin/server.rs
technosophos: I changed all of the `println!`s in server.rs to `warn!`s

src/invoice/verification.rs
technosophos: You mentioned you were gonna change this. So I will leave it as-is

src/provider/file/mod.rs
technosophos: That makes sense

PR #128, opened 2021-04-16
chore(*): Switches to the `tracing` crate and instruments all server things

thomastaylor312:
This adds _a lot_ more tracing and instrumentation on most of the major functions
used for the server. It also adds in tracing to the warp processing pipeline, which
should help with debugging. The client code still needs to be updated with more logs
and with instrumentation, but I will do that in a follow up.

Please note that if you think the logs are too verbose or you don't like the format,
tracing is very very tweakable, so we can iterate on this and change it


technosophos (approved):
LGTM

I'm not sure I know exactly when in my own code to use some of the tracing macros, but this definitely makes sense.

PR #130, opened 2021-04-20
chore(*): Added tracing to client

### LINKED ISSUE #129: Add client tracing
#128 adds support for tracing on all the server parts. We should also add tracing to the client parts

thomastaylor312:
Closes #129


technosophos (approved):
Just waiting for the GH Action to pass... but the code looks solid

PR #135, opened 2021-04-23
add Accept-type negotiation for JSON and TOML

### LINKED ISSUE #132: Support JSON format for requests and responses
Although TOML is great for hand-coding invoices, it's not as widely or as well supported as JSON for parsing or programmatic composition.

It would be convenient if the server supported `Accept: application/json` on invoice GETs and `Content-Type: application/json` on invoice POSTs.  This would allow JSON parsers and serialisers to be used on platforms where these are better maintained than TOML parsers (such as .NET).

technosophos:
Add content negotiation for JSON and TOML.

This adds support for using the `Accept` header to try to find the best format for sending results back. In addition, it accepts `invoice.json` content if the `content-type` is one of `text/json` or `application/json`.

Closes #132

### DIFF
--- a/src/server/filters.rs
+++ b/src/server/filters.rs
@@ -9,7 +9,7 @@ use tracing_futures::Instrument;
 use warp::reject::{custom, Reject, Rejection};
 use warp::Filter;
 
-use super::TOML_MIME_TYPE;
+use super::{JSON_MIME_TYPE, TOML_MIME_TYPE};
 use crate::authn::Authenticator;
 use crate::authz::Authorizer;
 
@@ -204,6 +204,21 @@ async fn parse_toml<T: DeserializeOwned + Send>(buf: impl warp::Buf) -> Result<T
         warn!("Failed to deserialize TOML file: {}", err);
         custom(BodyDeserializeError { cause: err.into() })
     })
+}
+
+pub fn json<T: DeserializeOwned + Send>() -> impl Filter<Extract = (T,), Error = Rejection> + Copy {
+    warp::filters::header::exact_ignore_case("Content-Type", JSON_MIME_TYPE)
+        .and(warp::body::aggregate())
+        .and_then(parse_json)
+}
+
+async fn parse_json<T: DeserializeOwned + Send>(buf: impl warp::Buf) -> Result<T, Rejection> {
+    debug!("Parsing as JSON");
+    let mut raw = Vec::new();
+    buf.reader()
+        .read_to_end(&mut raw)
+        .map_err(|err| custom(BodyDeserializeError { cause: err.into() }))?;
+    serde_json::from_slice(&raw).map_err(|err| custom(BodyDeserializeError { cause: err.into() }))
 }
 
 #[instrument(level = "trace", skip(err))]
--- a/src/server/filters.rs
+++ b/src/server/filters.rs
@@ -9,7 +9,7 @@ use tracing_futures::Instrument;
 use warp::reject::{custom, Reject, Rejection};
 use warp::Filter;
 
-use super::{JSON_MIME_TYPE, TOML_MIME_TYPE};
+use super::TOML_MIME_TYPE;
 use crate::authn::Authenticator;
 use crate::authz::Authorizer;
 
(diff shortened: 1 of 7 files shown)


src/server/filters.rs
```diff
@@ -203,6 +203,21 @@ async fn parse_toml<T: DeserializeOwned + Send>(buf: impl warp::Buf) -> Result<T
     toml::from_slice(&raw).map_err(|err| custom(BodyDeserializeError { cause: err.into() }))
 }
 
+pub fn json<T: DeserializeOwned + Send>() -> impl Filter<Extract = (T,), Error = Rejection> + Copy {
```
thomastaylor312, line 206: There is a [built in](https://docs.rs/warp/0.3.1/warp/filters/body/fn.json.html) filter for parsing JSON. Any reason we need our own custom one?

src/server/handlers.rs
```diff
@@ -20,9 +20,15 @@ pub mod v1 {
     pub async fn query_invoices<S: Search>(
         options: QueryOptions,
         index: S,
+        accept_header: Option<String>,
     ) -> Result<impl warp::Reply, Infallible> {
         let term = options.query.clone().unwrap_or_default();
         let version = options.version.clone().unwrap_or_default();
+        debug!(
+            "Querying with term {} and version{}",
+            term.clone(),
```
thomastaylor312, line 29: These should be added as structured fields rather than string interpolation:

```rust
debug!(%term, %version, "Querying index");
```

Same goes for most of the new `debug`s in this PR

src/server/reply.rs
```diff
@@ -1,37 +1,95 @@
...
+/// For example, `Accept: text/json, application/toml;q=0.9` will cause encoding to be in JSON.
+/// If no suitable content type is found, this will encode in application/toml, as that
+/// is the behavior described in the spec.
+pub fn serialized_data<T>(val: &T, accept: String) -> SerializedData
 where
     T: Serialize,
 {
-    Toml {
-        inner: toml::to_vec(val).map_err(|e| {
+    let supported = vec![
```
thomastaylor312, line 27: Nit: Consider breaking this out into a const slice to avoid string allocation every time

src/server/reply.rs
```diff
@@ -1,37 +1,95 @@
...
+    debug!("negotiated MIME is '{}'", final_mime);
+    SerializedData {
+        inner,
+        mime: final_mime.to_owned(),
     }
 }
 
-/// A JSON formatted reply.
-pub struct Toml {
+fn parse_accept<'a>(header: String) -> Vec<String> {
```
thomastaylor312, line 63: Nit: This doesn't need to be a `String` as `split` is a method from `&str`. It will avoid an extra allocation as well as you already clone the parts you need

thomastaylor312, line 63: Any reason for the lifetime here? Doesn't seem like it needs it

src/server/routes.rs
```diff
@@ -71,6 +73,22 @@ pub mod v1 {
...
+        ) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone
+        where
+            P: Provider + Clone + Send + Sync,
+        {
+            warp::path("_i")
+                .and(warp::path::end())
+                .and(warp::post())
+                .and(with_store(store))
+                .and(filters::json())
+                .and(warp::header::optional::<String>("accept"))
```
thomastaylor312, line 91: All these `accept` header passing things are...not fun (I think this was the stuff you mentioned offline), but I really don't think there is any other way to do this

thomastaylor312 (changes requested):

technosophos:
To capture this for a followup PR: We are thinking of changing the Accept handler code to work like this:

```rust
           warp::path("_r")
                .and(warp::path("missing"))
                .and(warp::path::tail())
                .and(warp::get())
                .and(with_store(store))
                .and(filters::create_serializer())
                .and(warp::header::optional::<String>("accept"))
                .and_then(get_missing)
        }
    }
}

trait BodySerializer {
    fn serialize<T>(&self, object: T) -> impl warp::Reply;
}

fn create_serializer() -> impl Filter<Extract = (T,), Error = std::convert::Infallible>
where
    T: BodySerializer,
{
    warp::header::optional::<String>("accept").and_then(|header| {
        match header {
            Some(h) => {
                match h {
                    "application/toml" => return Toml::new(),
                    "application/json" => return Json::new(),
                    _ => return Toml::new(), // or error
                }
            }
            None => return Toml::new(),
        }
    })
}

```

PR #142, opened 2021-05-21
fix(server): Respects RUST_LOG variable

thomastaylor312:
I found out that krustlet wasn't respecting it for the same reason, so
I added the change here as well


technosophos (approved):
LGTM

PR #144, opened 2021-05-27
add a convenience method for checking if the parcel is in the default group

technosophos:
This is helpful for higher-level usage like Wagi


technosophos:
Turns out this is unnecessary, since there is a higher-level `is_global_group()` function that does this. And I even wrote that function. 😊

PR #145, opened 2021-05-27
add a group_members function on invoice

technosophos:
This adds a `group_members(name)` function on the invoice so that it is easy to get all of the parcels that belong to a group.

Ported over from Wagi.

### DIFF
--- a/src/invoice/mod.rs
+++ b/src/invoice/mod.rs
@@ -99,7 +99,7 @@ impl Invoice {
 
     /// Check whether a group by this name is present.
     pub fn has_group(&self, name: &str) -> bool {
-        let empty = vec![];
+        let empty = Vec::with_capacity(0);
         self.group
             .as_ref()
             .unwrap_or(&empty)
--- a/src/invoice/mod.rs
+++ b/src/invoice/mod.rs
@@ -99,18 +99,18 @@ impl Invoice {
 
     /// Check whether a group by this name is present.
     pub fn has_group(&self, name: &str) -> bool {
+        let empty = vec![];
         self.group
-            .clone()
-            .unwrap_or_default()
+            .as_ref()
+            .unwrap_or(&empty)
             .iter()
-            .find(|g| g.name == name)
-            .is_some()
+            .any(|g| g.name == name)
     }
 
     /// Get all of the parcels on the given group.
     pub fn group_members(&self, name: &str) -> Vec<Parcel> {
         // If there is no such group, return early.
-        if self.has_group(name) {
+        if !self.has_group(name) {
             info!(name, "no such group");
             return vec![];
         }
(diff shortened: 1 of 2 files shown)


src/invoice/mod.rs
```diff
@@ -96,6 +96,18 @@ impl Invoice {
         version_compare(self.bindle.id.version(), requirement)
     }
 
+    /// Get all of the parcels on the given group.
+    pub fn group_members(&self, name: &str) -> Vec<Parcel> {
+        // TODO: This does not enforce that there is actually a group entry.
```
thomastaylor312, line 101: Did this need to be addressed before merge?

thomastaylor312 (approved):
LGTM, just the question about whether the TODO needed to be addressed beforehand

src/invoice/mod.rs
```diff
@@ -96,9 +97,24 @@ impl Invoice {
         version_compare(self.bindle.id.version(), requirement)
     }
 
+    /// Check whether a group by this name is present.
+    pub fn has_group(&self, name: &str) -> bool {
+        self.group
```
thomastaylor312, line 102: This can be shortened and avoid the clone by doing:

```rust
self.group.as_ref().unwrap_or_default().any(|g| g.name == name)
```

src/invoice/mod.rs
```diff
@@ -96,9 +97,24 @@ impl Invoice {
...
+            .iter()
+            .find(|g| g.name == name)
+            .is_some()
+    }
+
     /// Get all of the parcels on the given group.
     pub fn group_members(&self, name: &str) -> Vec<Parcel> {
-        // TODO: This does not enforce that there is actually a group entry.
+        // If there is no such group, return early.
+        if self.has_group(name) {
```
thomastaylor312, line 113: Should this be `!self.has_group(name)`?

thomastaylor312 (changes requested):

src/invoice/mod.rs
technosophos: Doing that returns a reference to the vec, which I don't think can be satisfied by either unwrap_or_default or unwrap_or_else. Also, I think you still need the iter().

src/invoice/mod.rs
technosophos: Good catch. The unit test had a bug, and so gave a false positive.

src/invoice/mod.rs
technosophos: Alright... came up with a half-way workaround that omits the clone.

src/invoice/mod.rs
```diff
@@ -96,6 +97,33 @@ impl Invoice {
         version_compare(self.bindle.id.version(), requirement)
     }
 
+    /// Check whether a group by this name is present.
+    pub fn has_group(&self, name: &str) -> bool {
+        let empty = vec![];
```
thomastaylor312, line 102: This can be `Vec::with_capacity(0)` to avoid an allocation but not necessary to fix

thomastaylor312 (approved):

PR #147, opened 2021-06-01
Option to negotiate HTTP/2 instead of assuming

itowlson:
By default, the Bindle client goes straight to HTTP/2 when talking to a Bindle server.  This is usually safe because the Bindle server speaks HTTP/2.  However, on certain occasions, certain rare and vexatious occasions, when the Bindle server is hidden behind a proxy or tunnel, a low, brutish proxy or tunnel that clings to savage, degraded protocols such as HTTP/1.1.  The Bindle server is perfectly happy being talked via such a proxy or tunnel, but the _client_ cannot establish communication because the oafish, ignorant proxy fails to comprehend its initial HTTP/2 greeting.

For these cases, these sad cases, it is useful to be able to have the client approach the proxy using HTTP/1.1 (and request an upgrade to HTTP/2).  In my sample of 1 tunnelling service, this worked 100% of the time.  This PR adds an option to do exactly that.

For the time being, I have not surfaced this option on the Bindle CLI; it is only on the Bindle client crate.  However, we could make it available e.g. via an environment variable, or even default the CLI to this behaviour as the performance benefit of bypassing negotiation may not be significant in an interactive environment.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -7,6 +7,7 @@ pub mod load;
 use std::convert::TryInto;
 use std::path::Path;
 
+use reqwest::ClientBuilder;
 use reqwest::header;
 use reqwest::Client as HttpClient;
 use reqwest::{Body, RequestBuilder, StatusCode};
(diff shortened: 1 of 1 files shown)


itowlson:
Or we could fix our public Bindle server, that's the only reason I'm having to faff around with tunnels anyway.  _grin_

src/client/mod.rs
```diff
@@ -33,12 +34,40 @@ pub struct Client {
...
+        Self {
+            http2_prior_knowledge: true,
+        }
+    }
+}
+
 impl Client {
-    /// Returns a new Client with the given URL. This URL should be the FQDN plus any namespacing
+    /// Returns a new Client with the given URL, configured using the default options.
+    /// This URL should be the FQDN plus any namespacing
```
thomastaylor312, line 57: Super Nit: Weird wrapping

thomastaylor312 (commented):
This seems fine to me! I do want to talk over the problem you were having with the public bindle server before approving so I understand the whole scope of the issue

thomastaylor312 (approved):

PR #149, opened 2021-06-03
removed the concept of empty groups from spec

technosophos:
Per a long discussion with @itowlson, we have decided that a parcel that is a member of no groups is a non-feature. It means that there are parcels that must be present on the server, but which are never pulled. At best, it is a way of hiding data that a user could be directed to pull directly (via `curl`, for example). At worst, we made it really easy to hide nefarious stuff on a Bindle server.

### DIFF
--- a/docs/invoice-spec.md
+++ b/docs/invoice-spec.md
@@ -144,7 +144,10 @@ Between groups and conditions, it is possible to build tree-like dependency stru
 
 By default, if no condition is provided, an item is a member of the "global" group, and is required.
 
-- `memberOf`: A list of groups that this parcel is a member of. When a `memberOf` clause is present, the parcel is removed from the default global group and placed into _just_ the groups listed in the `memberOf` clause. `memberOf = []` indicates that this parcel is a member of no groups (including the global group). It is an error if a parcel references a group that is undefined in the `[[group]]` list. (OPTIONAL)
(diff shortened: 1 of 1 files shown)


docs/invoice-spec.md
```diff
@@ -144,7 +144,10 @@ Between groups and conditions, it is possible to build tree-like dependency stru
 
 By default, if no condition is provided, an item is a member of the "global" group, and is required.
 
-- `memberOf`: A list of groups that this parcel is a member of. When a `memberOf` clause is present, the parcel is removed from the default global group and placed into _just_ the groups listed in the `memberOf` clause. `memberOf = []` indicates that this parcel is a member of no groups (including the global group). It is an error if a parcel references a group that is undefined in the `[[group]]` list. (OPTIONAL)
+- `memberOf`: A list of groups that this parcel is a member of. When a `memberOf` clause is present, the parcel is removed from the default global group and placed into _just_ the groups listed in the `memberOf` clause.
```
itowlson, line 147: Possibly picky (since you cover this in the very next paragraph) but may be worth writing:

```suggestion
- `memberOf`: A list of groups that this parcel is a member of. When a nonempty `memberOf` clause is present, the parcel is removed from the default global group and placed into _just_ the groups listed in the `memberOf` clause.
```

itowlson (approved):
One minor suggestion otherwise LGTM

PR #153, opened 2021-06-10
fix(server): Allows for proper parsing of charset

### LINKED ISSUE #131: Accept charset in Content-Type header
When creating an invoice, if you send a Content-Type of `application/toml; charset=...`, the server rejects it with 400 Bad Request, because it tries to exactly match the string `application/toml` and the additional clause confuses it.  However this is allowed by the HTTP spec and is sent by default by at least the .NET HttpClient/StringContent class.  The server should accept the charset declaration.

thomastaylor312:
Closes #131

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,8 +65,6 @@ ed25519-dalek = "1.0"
 base64 = "0.13"
 tracing = { version = "0.1", features = ["log"] }
 tracing-futures = "0.2"
-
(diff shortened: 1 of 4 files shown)


itowlson (approved):
Looks fine to me (though I don't know the code) - thanks!

radu-matei (approved):
LGTM

PR #155, opened 2021-06-10
fix(provider/file): Uses partfile for invoice and improves locking

thomastaylor312:
This has some small improvements to the new `PartFile` to make sure that it
takes a file "lock" right when it is created. It also refactored the `create_invoice`
method to use the new `PartFile`. I also added a test to at least give
us a sanity check that a double write doesn't cause an issue

### DIFF
--- a/src/provider/file/mod.rs
+++ b/src/provider/file/mod.rs
@@ -187,39 +187,11 @@ impl<T: crate::search::Search + Send + Sync> Provider for FileProvider<T> {
             debug!("Invoice being created already exists in storage");
             return Err(ProviderError::Exists);
         }
+
         // Create the part file to indicate that we are currently writing
-        let part = part_path(&dest).await?;
-        trace!(path = %part.display(), "Checking that a write is not currently in progress");
-        // Make sure we aren't already writing
-        if tokio::fs::metadata(&part)
-            .await
-            .map(|m| m.is_file())
-            .unwrap_or(false)
-        {
-            return Err(ProviderError::WriteInProgress);
-        }
-        debug!(
-            path = %part.display(),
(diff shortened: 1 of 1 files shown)


src/provider/file/mod.rs
```diff
@@ -835,4 +812,59 @@ mod test {
...
+            ),
+            store.create_parcel(
+                &scaffold.invoice.bindle.id,
+                &parcel.sha,
+                FramedRead::new(std::io::Cursor::new(parcel.data.clone()), BytesCodec::new()),
+            ),
+        );
+
+        // At least one should fail
+        assert!(
```
radu-matei, line 860: I really like the way this test is set up.

src/provider/file/mod.rs
```diff
@@ -602,22 +575,26 @@ impl PartFile {
...
     /// Moves the file to the configured final location, consuming the part file
-    async fn finalize(self) -> Result<()> {
+    async fn finalize(mut self) -> Result<()> {
         debug!(
             renamed_path = %self.final_location.display(),
             "Renaming part file for parcel"
         );
+
+        // Close the file handle to avoid any problems with unfinished IO operations
+        self.file.shutdown().await?;
```
radu-matei, line 596: Did the previous version not close the file here?

radu-matei (approved):

src/provider/file/mod.rs
thomastaylor312: In this PR I added the file, before this PR it was dropped inside of the `write_data` function, which cleans it up. But because I am doing a `rename` on it, I didn't want any problems because I still had the file handle around

PR #158, opened 2021-06-17
The Clone Wars: Attack on the Clones

thomastaylor312:
This PR consists of several mostly atomic commits that do some cleanup and small refactors that were much needed (it has been a while since we last cleaned up the code). Each commit explains specifically what was done and why, so I would recommend reviewing each commit separately when reviewing.

### DIFF
--- a/bin/server.rs
+++ b/bin/server.rs
@@ -281,9 +281,9 @@ where
     // MPB: The original version did an unwrap_or_default() on the read_to_string.
     // I removed this because I think we want an error to propogate if the file
     // cannot be read.
-    let raw_data = tokio::fs::read_to_string(&file)
+    let raw_data = tokio::fs::read(&file)
         .await
         .map_err(|e| anyhow::anyhow!("failed to read TOML file {}: {}", file.display(), e))?;
-    let res = toml::from_str::<T>(&raw_data)?;
+    let res = toml::from_slice::<T>(&raw_data)?;
     Ok(res)
 }
--- a/bin/server.rs
+++ b/bin/server.rs
@@ -143,7 +143,7 @@ async fn main() -> anyhow::Result<()> {
                 .get("keyring")
                 .map(|v| v.as_str().unwrap().parse().unwrap())
         })
-        .unwrap_or(PathBuf::from(default_config_dir().join("keyring.toml")));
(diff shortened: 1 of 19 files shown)


thomastaylor312:
Currently doing this as a draft while I finish a few things

bin/server.rs
```diff
@@ -153,10 +153,12 @@ async fn main() -> anyhow::Result<()> {
     // All other cases are considered errors worthy of failing.
     let keyring: KeyRing = match std::fs::metadata(&keyring_file) {
         Ok(md) if md.is_file() => load_toml(keyring_file).await?,
-        Ok(_) => Err(anyhow::anyhow!(
-            "Expected {} to be a regular file",
-            keyring_file.display()
-        ))?,
+        Ok(_) => {
+            return Err(anyhow::anyhow!(
```
technosophos, line 157: Can this be reduced to `anyhow::bail!`? I always forget to use `bail!()`.

src/invoice/mod.rs
```diff
@@ -117,21 +117,21 @@ impl Invoice {
...
+        let zero_vec = Vec::with_capacity(0);
         self.parcel
-            .clone()
-            .unwrap_or_default()
+            .as_ref()
+            .unwrap_or(&zero_vec)
             .iter()
             .filter(|p| p.member_of(name))
-            .map(|p| p.clone())
+            .cloned()
```
technosophos, line 126: TIL....

src/invoice/verification.rs
```diff
@@ -5,6 +5,17 @@ use tracing::{debug, info};
 
 use std::convert::TryInto;
 
+const GREEDY_VERIFICATION_ROLES: &[SignatureRole] = &[SignatureRole::Creator];
```
technosophos, line 8: Ah! You mentioned this! Yeah, that came out nicely.

technosophos (approved):

technosophos:
Suggested: Change the PR name to "The Clone Wars: Attack on the clones"

bin/server.rs
thomastaylor312: Oh yeah I forgot about bail. Will fix

PR #161, opened 2021-06-22
Reworked error output for ID and SemVer parse errors

### LINKED ISSUE #159: `InvalidId(InvalidId)`
When the client reports an invalid ID error, it should say what the ID is.

technosophos:
This results in error output that walks the nested errors and prints the root cause:

```console
$ bindle --server $BINDLE_HOST push foo
Error: Invalid id: InvalidId("foo")

Caused by:
    Invalid bindle ID foo. A bindle ID should be NAME/VERSION

$ bindle --server $BINDLE_HOST push foo/bar/dadsfas
Error: Invalid id: InvalidSemver("dadsfas")

Caused by:
    ID does not contain a valid semantic version (e.g. 1.2.3): dadsfas
```

Closes #159

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -31,7 +31,7 @@ async fn main() {
     if let Err(e) = run().await.map_err(|e| anyhow::Error::new(e)) {
         eprintln!("{}", e);
         for (i, cause) in e.chain().enumerate() {
-            // Skip the first message
+            // Skip the first message because it is printed above.
             if i > 0 {
                 if i == 1 {
                     eprintln!("\nError trace:");
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -14,7 +14,6 @@ use bindle::{
 };
 
 use clap::Clap;
-use futures::TryFutureExt;
 use sha2::Digest;
 use tokio::io::AsyncWriteExt;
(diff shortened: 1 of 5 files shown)


bin/client/main.rs
```diff
@@ -26,7 +26,24 @@ mod opts;
 use opts::*;
 
 #[tokio::main]
-async fn main() -> std::result::Result<(), ClientError> {
+async fn main() {
+    // Trap and format error messages using the proper value
+    if let Err(e) = run().await.map_err(|e| anyhow::Error::new(e)) {
+        eprintln!("{}", e);
+        for (i, cause) in e.chain().enumerate() {
+            // Skip the first message
```
thomastaylor312, line 34: Nit: I forgot to add to this to say we skip it because it is the same as the initial error that gets printed

thomastaylor312 (approved):
Looks good. One minor thing we forgot in a comment, but no need for another review

bin/client/main.rs
technosophos: Fixed

technosophos:
Okay, so I just tested, and hyper remains blissfully oblivious to SSL. So I will need to fix that. Might as well do that in this PR.

PR #162, opened 2021-06-23
docs(README): Updates the Using Bindle section to reflect reality

thomastaylor312:


technosophos:
At some point we need to switch Bindle to us XDG correctly. Some things are stored in XDG_CONFIG_HOME and others are in `.bindle`.

technosophos (approved):
Test failure appears to be a flake

PR #164, opened 2021-06-24
SSL/TLS configuration streamlining

technosophos:
This streamlines the TLS configuration and matches it to a similar PR for Wagi.

- `make serve-tls` runs the server with TLS
- `make gen-cert` generates a testing TLS certificate
- The CLI args and env vars are changed to match Wagi's
- Other CLI help text was cleaned up to clarify the difference between TLS keys and signing keys
- The docs are updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ BINDLE_LOG_LEVEL ?= debug
 BINDLE_ID ?= enterprise.com/warpcore/1.0.0
 BINDLE_IFACE ?= 127.0.0.1:8080
 MIME ?= "application/toml"
-CERT_NAME ?= "./ssl-example"
+CERT_NAME ?= ssl-example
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -36,3 +42,6 @@ build-server:
 build-client:
 	cargo build ${CLIENT_FEATURES} --bin ${CLIENT_BIN}
 
+.PHONY: gen-cert
+gen-cert:
```
thomastaylor312, line 46: Super nit (for bonus points!): Have an actual make target for the cert (not a `PHONY` one) so it generates it automatically if it doesn't exist when `make serve` is run

thomastaylor312 (approved):

Makefile
technosophos: That is a GREAT idea.

technosophos: Done.

PR #165, opened 2021-06-24
feat(server): Adds configurable verification strategy

### LINKED ISSUE #143: Signing: API server needs to load strategy
At some point, the API server should be able to load the signature verification strategy. For now, it defaults to GreedyVerification, which is the strategy that the specification recommends.

thomastaylor312:
You are now able to specify the verification strategy with a command line
flag or an item in the config file. This ended up being a bit more involved
than we initially thought due to the need to implement a custom deserialize
for the strategy

Closes #143

### DIFF
--- a/bin/server.rs
+++ b/bin/server.rs
@@ -19,7 +19,7 @@ Bindle is a technology for storing and retrieving aggregate applications.
 This program runs an HTTP frontend for a Bindle repository.
 "#;
 
-#[derive(Clap)]
+#[derive(Clap, serde::Deserialize, Default)]
 #[clap(name = "bindle-server", version = clap::crate_version!(), author = "DeisLabs at Microsoft Azure", about = DESCRIPTION)]
 struct Opts {
     #[clap(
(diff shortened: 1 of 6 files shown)


src/server/mod.rs
```diff
@@ -28,6 +28,7 @@ pub struct TlsConfig {
 /// Returns a future that runs a server until it receives a SIGINT to stop. If optional TLS
 /// configuration is given, the server will be configured to use TLS. Otherwise it will use plain
 /// HTTP
+// TODO: Change this to a builder instead as this is too many arguments
```
thomastaylor312, line 31: Please note: I am working on a follow up PR for this TODO. Didn't want to mix it with this PR

thomastaylor312: So making a builder that works with generics is fairly impossible because once you choose a type, it needs to stay that type the whole time. Just going to remove this todo and ignore the clippy lint

thomastaylor312: To clarify, all 5 of the generic parameters would need to be provided up front. We'd only be removing 3 things. So it probably isn't worth it

bin/server.rs
```diff
@@ -76,6 +76,14 @@ struct Opts {
         about = "location of the TOML file that holds the signing keys"
     )]
     signing_file: Option<PathBuf>,
+
+    #[clap(
+        name = "verification_strategy",
+        long = "strategy",
+        env = "BINDLE_VERIFICATION_STRATEGY",
+        about = "The verification strategy to use on the server. Must be one of: CreativeIntegrity, AuthoritativeIntegrity, GreedyVerification, ExhaustiveVerification, MultipleAttestation, MultipleAttestationGreedy. For either of the multiple attestation strategies, you can specify the roles using the following syntax: `MultipleAttestation[Creator, Approver]`"
```
technosophos, line 84: `MultipleAttestation[Roles..]` is pretty cool

src/invoice/verification.rs
```diff
@@ -48,6 +49,84 @@ impl Default for VerificationStrategy {
...
+impl FromStr for VerificationStrategy {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.is_empty() {
+            return Err("Cannot parse VerificationStrategy from an empty string");
+        }
+        // We are very flexible here, allowing for any casing and trimming leading or trailing
+        // whitespace
+        let normalized = s.trim().to_lowercase();
```
technosophos, line 63: I like the flexibility here.

technosophos (approved):

PR #166, opened 2021-06-28
added accessor methods for StandaloneRead

technosophos:
When pairing with @adamreese on https://github.com/deislabs/wagi/issues/79, I noticed that we had some missing features on `StandaloneRead`. This adds accessors to the invoice and parcels so that client users do not have to know anything about the file structure of a standalone bindle.

Also refactored the `Makefile` because I am impatient and want to run tests faster.

### DIFF
--- a/src/standalone/mod.rs
+++ b/src/standalone/mod.rs
@@ -3,8 +3,10 @@ use std::collections::HashMap;
 use std::convert::TryInto;
 use std::path::{Path, PathBuf};
 
+use tokio::fs::File;
 use tokio::io::{AsyncRead, AsyncWriteExt};
 use tokio_stream::{Stream, StreamExt};
+use tokio_util::codec::{BytesCodec, FramedRead};
 use tracing::{debug, info, instrument, trace};
 
 use crate::client::{Client, ClientError, Result};
@@ -13,7 +15,9 @@ use crate::Id;
 /// The name of the invoice file
 pub const INVOICE_FILE: &str = "invoice.toml";
 /// The name of the parcels directory
-pub const PARCEL_DIR: &str = "parcels/";
+pub use crate::provider::file::{PARCEL_DAT, PARCEL_DIRECTORY as PARCEL_DIR};
+
+//const PARCEL_DIR: &str = "parcels/";
 
 /// A struct containing paths to all of the key components of a standalone bundle
 pub struct StandaloneRead {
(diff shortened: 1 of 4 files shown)


src/standalone/mod.rs
```diff
@@ -13,7 +15,9 @@ use crate::Id;
 /// The name of the invoice file
 pub const INVOICE_FILE: &str = "invoice.toml";
 /// The name of the parcels directory
-pub const PARCEL_DIR: &str = "parcels/";
+pub use crate::provider::file::{PARCEL_DAT, PARCEL_DIRECTORY as PARCEL_DIR};
+
+//const PARCEL_DIR: &str = "parcels/";
```
thomastaylor312, line 20: Left over code

src/standalone/mod.rs
```diff
@@ -116,6 +120,46 @@ impl StandaloneRead {
             .collect::<Result<Vec<_>>>()?;
         Ok(())
     }
+
+    /// Retrive the invoice from the standalone bindle.
```
thomastaylor312, line 124: Nit (spelling): Retrieve

src/standalone/mod.rs
```diff
@@ -116,6 +120,46 @@ impl StandaloneRead {
...
+        let inv = toml::from_slice(&data)?;
+        Ok(inv)
+    }
+
+    /// Get the path to a parcel in a standalone bindle
+    pub fn parcel_data_path(&self, parcel_id: &str) -> PathBuf {
+        self.parcel_dir.join(format!("{}.dat", parcel_id))
+    }
+
+    pub async fn get_parcel(&self, parcel_id: &str) -> Result<Vec<u8>> {
```
thomastaylor312, line 141: Missing documentation

src/standalone/mod.rs
```diff
@@ -116,6 +120,46 @@ impl StandaloneRead {
...
+    pub async fn get_invoice(&self) -> Result<crate::invoice::Invoice> {
+        let data = tokio::fs::read(&self.invoice_file)
+            .await
+            .map_err(ClientError::Io)?;
+        let inv = toml::from_slice(&data)?;
+        Ok(inv)
+    }
+
+    /// Get the path to a parcel in a standalone bindle
+    pub fn parcel_data_path(&self, parcel_id: &str) -> PathBuf {
```
thomastaylor312, line 137: Should there be an `invoice_data_path` as well for those who need access to the file path?

src/standalone/mod.rs
```diff
@@ -190,6 +234,8 @@ impl StandaloneWrite {
 
     /// Writes the given invoice and `HashMap` of parcels (as readers). The key of the `HashMap`
     /// should be the SHA of the parcel
+    //
+    // The hash mapp should be the sha256 of the data as a key, and the data as the value.
```
thomastaylor312, line 238: These should be triple slashed: `///`

thomastaylor312: Nit (spelling): map

src/standalone/mod.rs
```diff
@@ -322,3 +368,105 @@ fn validate_shas<'a, T: Iterator<Item = &'a String>>(
...
+                origin: None,
+                feature: None,
+            },
+            conditions: None,
+        }]);
+        let mut parcels = HashMap::new();
+        parcels.insert(sha_string.clone(), parcel_data);
+
+        // Save
+        eprintln!("Writing to {}", dir.path().display());
```
thomastaylor312, line 420: Left over debug or intentional?

src/standalone/mod.rs
```diff
@@ -322,3 +368,105 @@ fn validate_shas<'a, T: Iterator<Item = &'a String>>(
...
+        // Load invoice
+        let inv2 = reader.get_invoice().await.expect("load invoice");
+        assert_eq!(
+            "standalone/roundtrip/1.0.0".to_string(),
+            inv2.bindle.id.to_string()
+        );
+
+        // Load parcel
+        let parcel_data2 = reader
+            .get_parcel(sha_string.as_str())
```
thomastaylor312, line 462: We should probably also test the stream method as well here

thomastaylor312 (changes requested):
Code looks good, mostly left over debug stuff and documentation

src/standalone/mod.rs
technosophos: That is a good question. There is an `invoice_file` public attribute on the struct that is the full path. Is there a good reason to also have that as a method?

My reasoning behind `parcel_data_path()` is that (a) it exists on the file provider, and (b) to construct the path you have to understand the structure of a standalone bindle.

thomastaylor312: I forgot that we had the path publicly available. That should be fine then

thomastaylor312 (approved):

PR #168, opened 2021-06-29
feat(proxy): Adds support for signing and verifying as a proxy

### LINKED ISSUE #141: Proxy: Sign bindles in get_yanked_invoice
Right now, the proxy signs invoices on creation. I _think_ the proxy probably should also sign (with a proxy record) any request that it gets from an upstream.

See #127

thomastaylor312:
As part of this, I realized that `Proxy` isn't the best thing to be using
for the client because it isn't truly a Proxy as it is acting as a provider
wrapper. This meant that we'd have a bunch of verification and signing code
you could turn off when using the proxy. Instead of doing that I just
did a `Provider` implementation for `Client`. This is a little less DRY, but
better separation of concerns and a cleaner API for `Proxy`

Closes #141

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -28,7 +28,7 @@ use opts::*;
 #[tokio::main]
 async fn main() {
     // Trap and format error messages using the proper value
-    if let Err(e) = run().await.map_err(|e| anyhow::Error::new(e)) {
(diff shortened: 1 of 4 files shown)


thomastaylor312:
superseded by #172

PR #174, opened 2021-06-30
fix(file): Ensures that windows part file actually locks

### LINKED ISSUE #173: File locking doesn't work on windows
As evidenced by the flakes we have in our windows tests, it looks like the file locking in the file provider doesn't work on windows. I think this could be solved by using the windows specific `OpenOptionsExt` and setting the [share mode](https://doc.rust-lang.org/1.53.0/std/os/windows/fs/trait.OpenOptionsExt.html#tymethod.share_mode) to 0

thomastaylor312:
Based on the documentation for the windows specific code, it looks like
the share mode is set to allow other processes to read/write/delete the
file. This removes those permissions so that it should stop a new file
from being simultaneously created.

Closes #173

### DIFF
--- a/src/provider/file/mod.rs
+++ b/src/provider/file/mod.rs
@@ -530,6 +530,29 @@ impl PartFile {
         {
             return Err(ProviderError::WriteInProgress);
         }
+        #[cfg(target_family = "windows")]
+        let file = match OpenOptions::new()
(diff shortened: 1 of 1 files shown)


technosophos (approved):
LGTM

PR #175, opened 2021-07-01
fix(signature): Ensures that a saved key has 0700 bits on unix

### LINKED ISSUE #115: Bug: create-key should make sure the secret key file mode is 0700
Right now, it doesn't check

thomastaylor312:
Closes #115

### DIFF
--- a/src/invoice/signature.rs
+++ b/src/invoice/signature.rs
@@ -3,6 +3,8 @@
 pub use ed25519_dalek::{Keypair, PublicKey, Signature as EdSignature, Signer};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use tokio::fs::OpenOptions;
+use tokio::io::AsyncWriteExt;
 use tracing::error;
 
 use std::convert::{TryFrom, TryInto};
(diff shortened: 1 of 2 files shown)


src/server/reply.rs
```diff
@@ -160,8 +160,6 @@ pub fn reply_from_error(
 
 #[cfg(test)]
 mod test {
-    use mime::JSON;
```
technosophos, line 163: I hate spurious automatic includes from VS Code.

technosophos (approved):

src/invoice/signature.rs
```diff
@@ -337,7 +339,25 @@ impl SecretKeyFile {
...
-        tokio::fs::write(dest, out).await?;
+        #[cfg(target_family = "unix")]
+        let mut file = OpenOptions::new()
+            .create_new(true)
+            .write(true)
+            .mode(0o700)
+            .open(dest)
+            .await?;
+
+        // TODO(thomastaylor312): Figure out what the proper permissions are on windows (probably
```
thomastaylor312, line 350: Created #176 to track this

PR #183, opened 2021-07-20
feat(*): Adds `cargo deny` to build pipeline

thomastaylor312:
This adds package auditing and `cargo deny` to the build workflow for
bindle. This should inform us if we are using any projects with security
advisories or ones we have explicitly denied (none right now). This will
also run a license check to make sure we are at least using licensed
projects. It can be tweaked to allow specific licenses in the future.

As part of this, I have addressed any changes it found

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,189 @@
+# This template contains all of the possible sections and their default values
+
+# Note that all fields that take a lint level have these possible values:
+# * deny - An error will be produced and the check will fail
+# * warn - A warning will be produced, but the check will not fail
+# * allow - No warning or error will be produced, though in some cases a note
+# will be
+
+# The values provided in this template are the default values that will be used
+# when any section or field is not specified in your own configuration
+
+# If 1 or more target triples (and optionally, target_features) are specified,
+# only the specified targets will be checked when running `cargo deny check`.
+# This means, if a particular package is only ever used as a target specific
+# dependency, such as, for example, the `nix` crate only being used via the
+# `target_family = "unix"` configuration, that only having windows targets in
+# this list would mean the nix crate, as well as any of its exclusive
+# dependencies not shared by any other crates, would be ignored, as the target
+# list here is effectively saying which targets you are building for.
+targets = [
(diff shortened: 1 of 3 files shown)


thomastaylor312:
Note: We will need to turn on the new checks as required once this merges

deny.toml
```diff
@@ -0,0 +1,187 @@
...
+    { name = "hashbrown", version = "^0.9" }
+]
+
+# This section is considered when running `cargo deny check sources`.
+# More documentation about the 'sources' section can be found here:
+# https://embarkstudios.github.io/cargo-deny/checks/sources/cfg.html
+[sources]
+# Lint level for what to happen when a crate from a crate registry that is not
+# in the allow list is encountered
+unknown-registry = "deny"
```
thomastaylor312, line 175: Given the signing/security parts of this crate, I decided it would be safer if we deny any unknown sources

deny.toml
```diff
@@ -0,0 +1,187 @@
...
+# List of explictly allowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.11 short identifier (+ optional exception)].
+allow = [
+    "LicenseRef-ring",
+]
+# List of explictly disallowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.11 short identifier (+ optional exception)].
+deny = [
```
technosophos, line 78: I know we also ban the following licenes:

- Everyone but Richard Stallman License
- Do Whatever The Fuck You Want License

I _believe_ that we might ban CC-0, but I am not sure.

technosophos (approved):
LGTM. Added a minor nit, but it's non-blocking

deny.toml
thomastaylor312: I added all 3 and learned about 3 new licenses today!

thomastaylor312:
The windows and mac tests flaked again, but not going to worry about running again as it passed before the change in the `deny.toml`

PR #184, opened 2021-07-21
Spec: describe annotations and reserved annotations

### LINKED ISSUE #182: Add standard labels or groups for `README` and `LICENSE` files
We know that tooling will need to be able to find licenses, and we suspect that tooling will benefit from having a standard place for a top-level document. So it has been proposed that we add a standard way of labeling README and LICENSE files.

See https://github.com/deislabs/hippofactory/issues/21

technosophos:
This describes label annotations, defines two reserved annotations, and clarifies invoice annotations.

Closes #182 


radu-matei (approved):
LGTM

PR #194, opened 2021-07-26
feat(server): Improves errors returned from proxies and signature issues

### LINKED ISSUE #193: Print better errors when there is a SignatureError
Right now the only error printed by a client is something like `failed signature check`. We should either manually implement `Display` or do something else so that it provides a full error description to the user similar to what we did [here](https://github.com/deislabs/bindle/blob/0f66cf520f5992b37e7ec15d0e78297c8ae47eaf/bin/client/main.rs#L30)

thomastaylor312:
Our current implementation just to_string'd things for nested errors. This
led to very confusing errors for the user. Now this un-nests stuff to get a
a better message to the user for client errors

Before:

```
Invalid request (status code 400): Some("failed signature check invoice")
```

After:

```
Invalid request (status code 400): Some("no suitable key for signing data")
```

Closes #193

### DIFF
--- a/src/server/reply.rs
+++ b/src/server/reply.rs
@@ -116,14 +116,12 @@ impl Reply for SerializedData {
 /// error = "bindle is yanked"
 /// ```
 pub fn into_reply(error: ProviderError) -> warp::reply::WithStatus<SerializedData> {
-    let mut error = error;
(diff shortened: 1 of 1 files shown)


technosophos (approved):
LGTM

PR #200, opened 2021-07-29
add print-key command

technosophos:
This prints the public key entry for a private key in a bindle client's `secrets_keys.toml` or a bindle server's `signing-keys.toml` file.

Example:

```console
$ bindle print-key
version = "1.0"

[[key]]
label = "Matt <example@example.com>"
roles = ["creator"]
key = "Qx6Q8d0K8sb0828S8hivB9ev+CWB+i1jqv596F3MbFQ="
labelSignature = "v6trXDbBKLl+9dXgmNlySajcPYHTeqwPy3Zy+PybD5qKGfPoMhkEp1pnkpSWHz5KT1F5yGxyUAvk4EbFmW7KDg=="
```

Note that you can use `--label` to filter to only keys whose label matches. This is approximately consistent with the way GPG works.

```console
$ bindle print-key --label Matt
version = "1.0"

[[key]]
label = "Matt <example@example.com>"
roles = ["creator"]
key = "Qx6Q8d0K8sb0828S8hivB9ev+CWB+i1jqv596F3MbFQ="
labelSignature = "v6trXDbBKLl+9dXgmNlySajcPYHTeqwPy3Zy+PybD5qKGfPoMhkEp1pnkpSWHz5KT1F5yGxyUAvk4EbFmW7KDg=="
```

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -1,3 +1,4 @@
+use std::convert::TryInto;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
@@ -7,6 +8,7 @@ use bindle::invoice::signature::{
 };
 use bindle::invoice::Invoice;
 use bindle::provider::ProviderError;
+use bindle::signature::KeyEntry;
 use bindle::standalone::{StandaloneRead, StandaloneWrite};
 use bindle::{
     cache::{Cache, DumbCache},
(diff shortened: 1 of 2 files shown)


technosophos:
See also #198. This outputs keys in the format described there.

technosophos:
Also, if no matching keys are found: 

```console
$ bindle --server http://localhost/v1 print-key --label foo
version = "1.0"
key = []
```

If a key on the secret keys file fails to decode, a warning is sent to STDERR, but the process will continue on. This is how `gpg` works, and it makes sense, because otherwise a single corrupted key could prevent you from being able to ascertain anything about your keyring.

thomastaylor312 (approved):

thomastaylor312:
This is the mac os test flake I've been seeing. Should be fine to merge

PR #201, opened 2021-07-29
fix(tests): Attempts to fix macos flake

thomastaylor312:
I believe that the slowness of the GH runners was creating issues when using
`cargo run` to run the server binary - even after previously building it. This
changes things to run the built binary directly instead, avoid the overhead of
"rebuilding" it. It also makes things more generic so people consuming test
utils are not required to have a specific binary name

### DIFF
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -6,6 +6,7 @@ use bindle::client::Client;
 use bindle::testing;
 
 const ENV_BINDLE_URL: &str = "BINDLE_URL";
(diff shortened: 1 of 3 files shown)


thomastaylor312:
Well, apparently this doesn't fix the flake, but the changes might still be useful. I'll leave it open in case others agree

technosophos (approved):
I prefer this change just because it cuts down on duplicate work

PR #205, opened 2021-07-30
exp(provider): Adds support for experimental embedded provider

thomastaylor312:
This PR adds support for a new embedded `Provider` backed by the `sled`
embedded database and with more efficient data encoding. This new
provider is behind a feature flag and is not compiled into `bindle-server`
by default. Tests have also been updated to check both providers, but
the client tests still only test with the `FileProvider`.

The motivation behind this is that the file system provider requires a bunch of
bespoke logic to check things like locking and various OS specific things. We
also have to front everything with an in-memory LRU cache. Tools like `sled`
already have most of these features built in, plus additional guarantees from
being a database (such as atomic transations and upsert commands). This also
uses a binary encoding format to store the data using less bytes and to
decrease serialization and deserialization time. All of these reasons
sounded like pretty good justification for giving it a whirl. Use of bindle
can continue as before until we decide whether we like this or not as it
is all behind a feature flag.

Any comments on open TODOs are welcome!

### DIFF
--- a/bin/server.rs
+++ b/bin/server.rs
@@ -192,7 +192,6 @@ async fn main() -> anyhow::Result<()> {
     tracing::info!("Using verification strategy of {:?}", strategy);
 
     let index = search::StrictEngine::default();
-    let store = provider::file::FileProvider::new(&bindle_directory, index.clone()).await;
     let secret_store = SecretKeyFile::load_file(&signing_keys).await.map_err(|e| {
         anyhow::anyhow!(
             "Failed to load secret key file from {}: {} HINT: Try the flag --signing-keys",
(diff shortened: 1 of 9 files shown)


bin/server.rs
```diff
@@ -201,24 +200,55 @@ async fn main() -> anyhow::Result<()> {
...
-        bindle::authn::always::AlwaysAuthenticate,
-        bindle::authz::always::AlwaysAuthorize,
-        addr,
-        tls,
-        secret_store,
-        strategy,
-        keyring,
-    )
-    .await
+    #[cfg(not(feature = "embedded"))]
```
technosophos, line 203: I don't understand why we would decide this at compile time instead of at runtime. Wouldn't this require compiling multiple binary distributions of Bindle?

src/provider/embedded.rs
```diff
@@ -0,0 +1,462 @@
...
+///
+/// Given a storage directory, EmbeddedProvider brings its own storage layout for keeping track of
+/// Bindles.
+///
+/// An EmbeddedProvider needs a search engine implementation. When invoices are created or yanked,
+/// the index will be updated.
+pub struct EmbeddedProvider<T> {
+    invoices: sled::Tree,
+    parcels: sled::Tree,
+    index: T,
```
technosophos, line 50: This is interesting... we could actually implement search indices in Sled as well, which would mean we could persist search indices. For full (non-strict) searches, we could implement something really robust! This would be great!

src/provider/embedded.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// This warms the index by loading all of the invoices currently in the DB
+    ///
+    /// Warming the index is something that the storage backend should do, though I am
+    /// not sure whether EVERY storage backend should do it. It is the responsibility of
+    /// storage because storage is the sole authority about what documents are actually
+    /// in the repository. So it needs to communicate (on startup) what documents it knows
+    /// about. The storage engine merely needs to store any non-duplicates. So we can
+    /// safely insert, but ignore errors that come back because of duplicate entries.
+    #[instrument(level = "trace", skip(self))]
+    async fn warm_index(&self) -> anyhow::Result<()> {
```
technosophos, line 96: I think this is still needed?

technosophos:
TBH, I like this method a lot more than the file method.

The only thing I can think of that makes the file method nice is that it is fairly easy for different languages to use the same "cache" of Bindles. So perhaps there is some value in keeping that as a documented option. But for an actual Bindle server, this storage method seems far preferable. We should be able to pull of much better performance while also doing less boilerplate IO.

I think most of our client implementations are using the standalone bindle format. (The early stuff, like cargo2bindle and npm2bindle were not). We should probably check on the other Bindle client libraries to see what they do. But again... simply allowing a user to specify whether they want the embedded version or the file-backed version is probably all we need to commit to.

bin/server.rs
thomastaylor312: Because it is experimental. I didn’t want to expose by default

src/provider/embedded.rs
thomastaylor312: Oh I forgot to implement this.

thomastaylor312:
Yeah if you are ok with that I can compile in both and then allow the user to select with a flag

thomastaylor312:
Oh and we can still use the file provider and standalone stuff just fine here for the CLI. No changes needed there

bin/server.rs
```diff
@@ -84,6 +84,15 @@ struct Opts {
         about = "The verification strategy to use on the server. Must be one of: CreativeIntegrity, AuthoritativeIntegrity, GreedyVerification, ExhaustiveVerification, MultipleAttestation, MultipleAttestationGreedy. For either of the multiple attestation strategies, you can specify the roles using the following syntax: `MultipleAttestation[Creator, Approver]`"
     )]
     verification_strategy: Option<bindle::VerificationStrategy>,
+
+    #[clap(
```
technosophos, line 88: ❤️

src/provider/embedded.rs
technosophos: Ah, this new one looks great.

technosophos (approved):
LGTM

PR #210, opened 2021-08-04
Provide better errors for non-GET operations

technosophos:
All 404s were getting interpreted as missing index/parcel. Turns out some 404s occur because the endpoint is missing. So this at least tries to distinguish between the two situations.

### DIFF
--- a/src/client/error.rs
+++ b/src/client/error.rs
@@ -34,6 +34,9 @@ pub enum ClientError {
     /// The parcel was not found.
     #[error("Parcel was not found")]
     ParcelNotFound,
+
+    #[error("Requested resource or endpoint is not found")]
+    ResourceNotFound,
     /// The invoice already exists
     #[error("Invoice already exists")]
     InvoiceAlreadyExists,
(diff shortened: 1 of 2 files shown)


technosophos:
I am not sure if the Windows failure was a flake or a legit failure. Lemme look into that

technosophos:
Here's the new output when hitting an endpoint that does not exist on the remote server:

```console
$ bindle --server http://example.com/v1 push-invoice _scratch/invoice.toml 
Requested resource or endpoint is not found
```

That error only shows up when there is a 404 error on an operation that is NOT a `GET` on a parcel or invoice.

technosophos:
I updated the output to include the URL:

```console
$ bindle --server http://example.com/v1 push-invoice _scratch/invoice.toml
Requested resource or endpoint is not found at URL http://example.com/v1/_i
```

thomastaylor312 (approved):

itowlson:
Seeing the endpoint makes this much clearer, at least to developers/people debugging an issue.  Nice one - thank you!

PR #213, opened 2021-08-10
feat(server): Adds Github authentication support to the server

### LINKED ISSUE #209: Implement OAuth2 workflow for Bindle
Attempt to integrate this, and ideally be able to OAuth against GitHub OAuth

thomastaylor312:
So this is a bit gnarly here, but functional. Basically, the login flow goes like this:

Client calls `/login` endpoint with their desired provider -> bindle server starts device code flow ->
bindle server returns auth code details to client -> client calls proper endpoint to retrieve token

On every API call, the passed token will be validated against the auth provider. We will likely
want to optimize this with some sort of cache in the future. Ideally, we can add other providers
in the future using OIDC.

Closes #209

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,10 +73,11 @@ dependencies = [
  "lru",
  "mime",
  "mime_guess",
+ "oauth2",
  "rand 0.7.3",
  "reqwest",
  "rstest",
- "semver",
+ "semver 0.11.0",
  "serde",
  "serde_cbor",
  "serde_json",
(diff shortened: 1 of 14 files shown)


thomastaylor312:
Oh, I should note: In order to test this, you need to create a github oauth application at https://github.com/settings/developers

technosophos (approved):
Finally finished reviewing. I think it looks good to merge... though hopefully we will figure out a more elegant method than the Match Statement of Doom.

PR #218, opened 2021-08-19
add HTTP Basic Auth to the client and the server

### LINKED ISSUE #217: Add HTTP Basic auth
Right now, the only auth we have is OAuth2, which is definitely preferred for production. But HTTP Basic might be a nice "lowest common denominator" auth scheme.

technosophos:
This adds support for HTTP Basic Auth. Per the docs/README.md:

To use HTTP Basic authentication, you must generate an `htpasswd` file using Bcrypt:

```console
$ htpasswd -Bc htpasswd admin
New password: 
Re-type new password: 
Adding password for user admin
```

(The `htpasswd` program comes on many Linux/Unix distros. Officially it is part of the Apache Web Server project.)

Then you need to supply the `--htpasswd-file` option at `bindle-server` startup:

```console
$ bindle-server --htpasswd-file test/data/htpasswd
```

> Currently, only bcrypt is supported in htpasswd files. At the time of this writing, bcrypt is the most secure algorithm supported by htpasswd.

Closes #217

### DIFF
--- a/bin/client/opts.rs
+++ b/bin/client/opts.rs
@@ -42,6 +42,16 @@ pub struct Opts {
     )]
     pub token_file: Option<PathBuf>,
 
+    #[clap(
+        long = "http-user",
+        about = "Username for HTTP Basic auth",
+        requires = "http-password"
+    )]
+    pub http_user: Option<String>,
+
+    #[clap(long = "http-password", about = "Password for HTTP Basic auth")]
+    pub http_password: Option<String>,
+
     #[clap(subcommand)]
     pub subcmd: SubCommand,
 }
--- a/src/authn/http_basic.rs
+++ b/src/authn/http_basic.rs
@@ -1,0 +1,139 @@
+use std::{collections::HashMap, path::Path};
+
+use super::Authenticator;
+use crate::authz::always::Anonymous;
+
+/// HTTP header prefix
+const HTTP_BASIC_PREFIX: &str = "Basic ";
+
+/// An authenticator that simply returns an anonymous user
+///
+/// In basic auth, the auth_data will come in as 'Basic BASE64_STRING', where
+/// the Base-64 string is the username and password separated by a colon.
+///
+/// This tool splits username and password, looks up the user in the database,
+/// and then compares the hashed password to the hash returned by the database.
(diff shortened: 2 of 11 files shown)


technosophos:
The failure on Windows has something to do with bcrypt and maybe whitespace. So I will look into that.

technosophos:
Okay, this is ready for review

bin/client/opts.rs
```diff
@@ -42,6 +42,16 @@ pub struct Opts {
...
     pub token_file: Option<PathBuf>,
 
+    #[clap(
+        long = "http-user",
+        about = "Username for HTTP Basic auth",
+        requires = "http-password"
+    )]
+    pub http_user: Option<String>,
+
+    #[clap(long = "http-password", about = "Password for HTTP Basic auth")]
```
thomastaylor312, line 52: This should require `http-user` as well. Also, I'd recommend setting `env` on both of these so the password isn't passed on the command line

src/authn/http_basic.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+fn parse_basic(auth_data: &str) -> anyhow::Result<(String, String)> {
+    match auth_data.strip_prefix(HTTP_BASIC_PREFIX) {
+        None => anyhow::bail!("Wrong auth type. Only Basic auth is supported"),
+        Some(suffix) => {
+            // suffix should be base64 string
+            let decoded = String::from_utf8(base64::decode(suffix)?)?;
+            let pair: Vec<&str> = decoded.splitn(2, ':').collect();
+            Ok((pair[0].to_owned(), pair[1].to_owned()))
```
thomastaylor312, line 103: We should validate length above here to avoid a DOS attack. If I send a base64 encoded string with no `:` separator, `pair[1]` will panic

thomastaylor312 (changes requested):

bin/client/opts.rs
technosophos: fixed

src/authn/http_basic.rs
technosophos: fixed

PR #221, opened 2021-08-26
Box Warp filters to speed up build time

technosophos:
This boxes the warp filters so that the compile-time checks do not take 5-15 minutes.


adamreese (approved):
Needs a refactor after adding the /health endpoint

PR #224, opened 2021-09-02
chore(*): Remove file provider option from server

thomastaylor312:
Also fixes one minor warning when only building with the `server` feature enabled


technosophos:
I was just telling Ivan that maybe we didn't want to make this change after all if we don't have the necessary time to do any load/restore/backup tools. Should we hold onto it for a few days and discuss early next week?

thomastaylor312:
Yeah we can hold for now!

PR #229, opened 2021-09-09
fix(makefile): Adds missing info for flag

thomastaylor312:
The change to make the embedded db flag take an optional now forces
you to pass an actual `bool` value to the flag


adamreese (approved):

thomastaylor312:
Superseded by #233

PR #233, opened 2021-09-10
Don't require a value passed for the --use-embedded-db

adamreese:


thomastaylor312 (changes requested):
Can you also update the makefile to remove the explicit `true` from the flag?

adamreese:
That PR was not merged

thomastaylor312:
Ah right. I lost track

thomastaylor312 (approved):

PR #236, opened 2021-09-16
Upgrade deps with `cargo upgrade`.

peterhuene:
This commit upgrades bindle's deps with `cargo upgrade`.

For almost all the deps, this just sets the patch version to the latest available, thus treating it as the minimum patch level for the given major.minor.

This specifically upgrades the `semver` crate to latest, as bindle was using a very old version that prevents downstream crates from using a newer version. As a result, a code change was required to maintain "npm" compatibility (which is no longer a thing in the `semver` crate).

The `dirs` create is upgraded to `4.0` with this change (API compatible).

The `rand` dependency was not updated as an upstream dependency uses a
backwards incompatible version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,9 +33,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.43"
+version = "1.0.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


peterhuene:
I'm also happy to just limit this change to bumping the `semver` crate version rather than running `cargo upgrade` (that's needed for my downstream use; `cargo upgrade` just makes it easier to bump minimum versions of things).

Issue #235, opened 2021-09-13
Bindle server should require an explicit flag to start without write auth


simongdavies:
Currently if I run bindle-server the write endpoints are unauthenticated, this should require explicit configuration

PR #237, opened 2021-09-16
Add unauthenticated flag to bindle-server

adamreese:
Makes running bindle-server without authentication an explicit option.

Closes: #235


thomastaylor312 (approved):

bacongobbler (approved):
--unauthenticated seems like a clear option.

Though should we recommend the user attempt to set up authentication? If we recommend users run `bindle-server --unauthenticated` then I feel users will immediately gravitate towards that option.

Either way, LGTM.

PR #238, opened 2021-09-16
Refactor merging config with cli options

adamreese:


fibonacci1729 (approved):

technosophos (approved):
I like this quite a bit.

Issue #234, opened 2021-09-13
Bindle info doesn't have --output 


squillace:
Unlike Bindle search, bindle info does not have any output formatting. 

```
╭─squillace at dellsquill in ~/work/squillace
╰─○ bindle info mybindle/0.1.0
bindleVersion = "1.0.0"

[bindle]
name = "mybindle"
version = "0.1.0"
description = "My first bindle"
authors = ["Matt Butcher <matt.butcher@microsoft.com>"]

[annotations]
myname = "myvalue"

[[signature]]
by = "Default host key"
signature = "suU7iEamliUEykecMT7ioXRIOrkPJVJtM0HKumLB66DjlA6SH7sQHrSiNGGKCtX5+qNbooAgzfGqDcaLBDX7Cg=="
key = "heEsdULqcm8/J//Lc/Io+AuRkKvI+ixIoaTlVZCPIhw="
role = "host"
at = 1631568557
```

PR #242, opened 2021-09-20
feat(client): add json output to bindle info

adamreese:
Closes: #234

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -144,7 +144,21 @@ async fn run() -> std::result::Result<(), ClientError> {
             }
             .await
             .map_err(map_storage_error)?;
-            tokio::io::stdout().write_all(&toml::to_vec(&inv)?).await?;
+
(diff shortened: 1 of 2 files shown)


bin/client/main.rs
```diff
@@ -144,7 +144,21 @@ async fn run() -> std::result::Result<(), ClientError> {
             }
             .await
             .map_err(map_storage_error)?;
-            tokio::io::stdout().write_all(&toml::to_vec(&inv)?).await?;
+
+            match info_opts.output {
+                Some(format) if &format == "toml" => {
```
radu-matei, line 149: non-blocking nit: this could probably be a match on the `Option` first, then a match for the string against string literals for the output.

radu-matei (approved):

bin/client/main.rs
adamreese: I'll refactor this in a different PR because it's in multiple places

PR #247, opened 2021-09-23
chore(bin): remove cargo2bindle and as2bindle

adamreese:


radu-matei:
Why are we removing them?

technosophos:
I don't know how up-to-date they are. I haven't touched them in a while... but if we want to move them to examples/, that is fine with me. But I agree with @adamreese that they should not be on equal footing with the bindle client and server.

adamreese:
Closing in favor of https://github.com/deislabs/bindle/pull/248

PR #248, opened 2021-09-27
chore(bin): move cargo2bindle and as2bindle to examples

adamreese:


radu-matei (approved):

thomastaylor312:
The test failures looks like they were flakes. The advisory can be ignored for now, but we should run a `cargo update` to pick up transitive dependency updates that should resolve the issue

PR #252, opened 2021-10-06
Emit events for data changes

### LINKED ISSUE #180: Add Change notification to bindle server
Bindle server should support event notification for changes, clients should be able to subscribe to an event stream, messages, webhook to get notifications about new bindles, new versions etc.

This will enable many scenarios including application updates, slack notifications etc.

simongdavies:
closes #180 

This change enables bindle server to emit events describing changes that have occurred to the data it manages, the feature is enabled by running the server with the flag `--events`.  Enabling the feature causes bindle to serialize details of changes in JSON format to a file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ openid = { version = "0.9.3", optional = true }
 bcrypt = "0.10.1"
 chrono = { version = "0.4.19", features = ["serde"], optional = true }
 either = "1.6.1"
+async-std = "1.9.0"
 
 # NOTE: This is a workaround due to a dependency issue in oauth2: https://github.com/tkaitchuck/ahash/issues/95#issuecomment-903560879
 indexmap = "~1.6.2"
--- a/docs/README.md
+++ b/docs/README.md
@@ -214,6 +214,29 @@ This file can be moved from system to system, just like OpenPGP or SSH key sets.
 - To create a signing key for a client, use `bindle create-key`
 - By default, if Bindle does not find an existing keyring, it creates one of these when it first starts.
 
+## Generating Change Events
+
+Starting bindle with the `--events`  flag will cause it to emit an record of the event to a file named `bindle-event.log` in the bindle data directory.
+
+Each event is serialized as a JSON object in the following format:
+``` json
+{
+  "event_date": "2021-10-05T20:05:51.028318270Z",
+  "event_data": {
+    "EventType": EventSpecificData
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -75,6 +75,7 @@ openid = { version = "0.9.3", optional = true }
 bcrypt = "0.10.1"
 chrono = { version = "0.4.19", features = ["serde"], optional = true }
 either = "1.6.1"
+async-std = "1.9.0"
```
thomastaylor312, line 78: I think this is going to require `features = ["tokio1"]` so that we aren't running another runtime alongside Tokio

docs/README.md
```diff
@@ -214,6 +214,29 @@ This file can be moved from system to system, just like OpenPGP or SSH key sets.
 - To create a signing key for a client, use `bindle create-key`
 - By default, if Bindle does not find an existing keyring, it creates one of these when it first starts.
 
+## Generating Change Events
+
+Starting bindle with the `--events`  flag will cause it to emit an record of the event to a file named `bindle-event.log` in the bindle data directory.
```
thomastaylor312, line 219: We might want to have a note about using some sort of tool like `logrotate` (or we should provide one) to avoid getting an infinitely growing file

src/events/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+pub trait EventSink {
+    // Emits an event.
+    async fn raise_event<T>(&self, _: EventType<T>) -> Result<(), Box<dyn std::error::Error>>
+    where
+        T: Serialize + Send + Sync;
+}
+
+// This enum exists to enable the type of event sync to be dynamically selected at runtime depending on arguments passed
+// to the program.
+// It seems like the normal approach of using trait objects does not work as the trait is not object safe as it has a generic type parameter.
```
thomastaylor312, line 26: Yep, you are correct here. It is the one glaring weakness of the Trait system. Generally we have not been exposing these publicly and have been putting them in the binary (as anyone can technically write something to implement the trait. Not going to block on this as we are still pre-1.0 and can break this later if needed

src/events/filesink.rs
```diff
@@ -0,0 +1,112 @@
+use async_std::fs::OpenOptions;
```
thomastaylor312, line 1: Actually, now that I am looking at this again, I think all of these options are available in tokio (including the prelude below)

PR #256, opened 2021-10-21
chore(*): Bumps bindle to use the 2021 edition

thomastaylor312:
We've been tracking and using most of the latest features anyway, so
this seems to make sense

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -35,13 +35,12 @@ jobs:
           OS=$(echo "${{ runner.os }}" | tr '[:upper:]' '[:lower:]')
           echo "RUNNER_OS=$OS" >> $GITHUB_ENV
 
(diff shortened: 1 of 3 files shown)


thomastaylor312:
NOTE: This cannot be merged until Github actions updates to 1.56

radu-matei:
We could manually `rustup install` and `default` to 1.56 in the meantime, if we wanted to.

thomastaylor312:
Yeah, we could use actions-rs to install the right version

thomastaylor312:
@radu-matei This is ready for re-review. The cargo deny stuff is handled in another PR

technosophos (approved):
It appears that GH Actions are now at the right version (see comments). So I think this is good to merge?

radu-matei (approved):
LGTM

PR #260, opened 2021-11-19
feat(cli): only use ansi logging on tty

adamreese:


bin/client/main.rs
```diff
@@ -75,7 +75,9 @@ impl TokenManager for PickYourAuth {
 async fn run() -> std::result::Result<(), ClientError> {
     let opts = opts::Opts::parse();
     // TODO: Allow log level setting outside of RUST_LOG (this is easier with this subscriber)
-    tracing_subscriber::fmt::init();
+    tracing_subscriber::fmt()
+        .with_ansi(atty::is(atty::Stream::Stderr))
```
thomastaylor312, line 79: I just realized we didn’t change the client to use stderr for logging. Should just be able to copy paste from server

thomastaylor312 (changes requested):
One minor change needed

bin/client/main.rs
technosophos: You mean essentially this?

https://github.com/deislabs/bindle/blob/main/bin/server.rs#L157-L160

thomastaylor312: Yep, exactly!

adamreese: @thomastaylor312 Can you check to make sure I understood correctly?

thomastaylor312: Yep, that is correct!

thomastaylor312 (approved):

PR #261, opened 2021-12-07
Adds codeowners

thomastaylor312:
This will make it so we get auto assigned on reviews so we don't miss them among our other notifications (like I did 🤦 ). @technosophos and @bacongobbler I added both of you. If there are others who we need to add, please do so!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @thomastaylor312 @technosophos @radu-matei


technosophos (approved):
LGTM

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @thomastaylor312 @technosophos @bacongobbler
```
radu-matei, line 1: Chatting with @bacongobbler, he might not be necessarily interested in being a maintainer.
Feel free to add me instead, as I plan to allocate some more time to work here.

thomastaylor312: Ok, I'll switch that out

bacongobbler (approved):

PR #262, opened 2021-12-07
Fixes cargo deny and update dependencies

thomastaylor312:
This updates a bunch of dependencies that were causing cargo deny issues and other compilation problems. Please note that I still had to ignore the omnipresent `time` and `chrono` RUSTSEC advisories as `openid` has not updated. I plan on submitting a PR to use the `time` crate instead over there to fix the issue and then those could be removed. However, I did address all other direct and transitive imports of `chrono`

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -26,15 +26,6 @@ jobs:
       - name: Run tests
         run: make test
 
-  security_audit:
-    name: Run security audit
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2.3.4
-      - uses: actions-rs/audit-check@v1.2.0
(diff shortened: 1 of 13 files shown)


thomastaylor312:
Please note that I removed the cargo audit step as it appears to be doing the exact same thing as `cargo deny` in this case. We may want to consider adding it back eventually as its own separate pipeline with a scheduled trigger

technosophos (approved):
Code LGTM

The Windows build failed, but apparently on a test that looks like maybe a race condition in CI?

thomastaylor312:
Yeah, that test still flakes every once and a while

radu-matei (approved):
Thank you so much for handling this, @thomastaylor312!

PR #283, opened 2022-01-27
feat(standalone): Adds tarball support to standalone bindles

thomastaylor312:
This also updates the CLI to take advantage of this new feature. Namely, the `get` command when using export will now output a tarball rather
than a directory. The `push` command now can use a directory or a tarball and
will infer that automatically from the file extension. I also added a new
`package` command that will package up a standalone bindle directory as a
tarball. An `expand` command was considered but rejected for the reason that
there is no special tooling required outside of `tar`. If we wish to add this
in the future, that is totally fine


thomastaylor312:
Labeled this as breaking as the CLI now exports bindles as tarballs when using `bindle get` so we make sure to capture the change in release notes

src/standalone/mod.rs
```diff
@@ -482,4 +588,97 @@ mod test {
         // Otherwise, tmpfile will clean up the tmpdir too soon.
         dir.close().expect("deleted temp dir");
     }
+
+    #[tokio::test]
+    async fn should_round_trip_tarball() {
```
technosophos, line 593: Nice test! This was super easy to follow as a guide to how things should work

technosophos (approved):

PR #294, opened 2022-02-02
feat(*): Adds required verification to the client

thomastaylor312:
The default strategy is just to check that the host signature is valid, so
this shouldn't provide too much of a burden for people kicking the tires

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -7,6 +7,7 @@ pub mod tokens;
 
 use std::convert::TryInto;
 use std::path::Path;
+use std::sync::Arc;
 
 use reqwest::header::{self, HeaderMap};
 use reqwest::Client as HttpClient;
@@ -16,8 +17,9 @@ use tracing::{debug, info, instrument, trace};
 use url::Url;
 
 use crate::provider::{Provider, ProviderError};
-use crate::verification::Verified;
-use crate::{Id, Signed};
+use crate::signature::{KeyRing, SignatureRole};
+use crate::verification::{Verified, VerifiedInvoice};
+use crate::{Id, Invoice, Signed, VerificationStrategy};
 
 pub use error::ClientError;
 
@@ -36,6 +38,9 @@ pub struct Client<T> {
     client: HttpClient,
     base_url: Url,
     token_manager: T,
+    verification_strategy: VerificationStrategy,
+    // In an arc as the keyring could be fairly large and we don't want to clone it everywhere
+    keyring: Arc<KeyRing>,
 }
 
 /// The operation being performed against a Bindle server.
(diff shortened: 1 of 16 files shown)


src/invoice/signature.rs
```diff
@@ -165,6 +165,7 @@ impl<T: AsRef<Path> + Sync> KeyRingSaver for T {
             .await?;
 
         file.write_all(&toml::to_vec(keyring)?).await?;
+        file.flush().await?;
```
technosophos, line 168: Ah, yes...

src/invoice/mod.rs
```diff
@@ -455,8 +455,8 @@ mod test {
             .expect("If no signature, then this should verify fine");
 
         // Create two signing keys.
-        let signer_name1 = "Matt Butcher <matt@example.com>".to_owned();
-        let signer_name2 = "Not Matt Butcher <not.matt@example.com>".to_owned();
+        let signer_name1 = "Matt Butcher <matt@example.com>";
+        let signer_name2 = "Not Matt Butcher <not.matt@example.com>";
```
technosophos, line 459: I feel oddly relieved that neither me nor not me is owned anymore.

src/client/mod.rs
```diff
@@ -48,10 +53,22 @@ enum Operation {
...
     danger_accept_invalid_certs: bool,
+    verification_strategy: VerificationStrategy,
+}
+
+impl Default for ClientBuilder {
+    fn default() -> Self {
+        ClientBuilder {
+            http2_prior_knowledge: false,
+            danger_accept_invalid_certs: false,
+            verification_strategy: VerificationStrategy::MultipleAttestationGreedy(vec![
```
technosophos, line 67: This seems like the sensible default.

src/client/mod.rs
```diff
@@ -113,8 +150,11 @@ impl<T: tokens::TokenManager> Client<T> {
     /// This URL should be the FQDN plus any namespacing (like `v1`). So if you were running a
     /// bindle server mounted at the v1 endpoint, your URL would look something like
     /// `http://my.bindle.com/v1/`. Will return an error if the URL is not valid
-    pub fn new(base_url: &str, token_manager: T) -> Result<Self> {
-        ClientBuilder::default().build(base_url, token_manager)
+    ///
+    /// This requires `KeyRing` to be wrapped in an `Arc` for efficiency sake. `KeyRing`s can be
+    /// quite large and so we require it to be wrapped in an Arc to avoid a large clone cost
```
technosophos, line 155: I was gonna ask why you used an Arc. This makes total sense.

technosophos (approved):

src/client/mod.rs
thomastaylor312: Yeah I figured it would be best to expose this directly in the API as an `Arc` because there are other places you could be needing the same keyring in code. This accounts for that so you can clone it willy nilly if you so desire

PR #295, opened 2022-02-03
docs(spec): Adds new Keyring protocol spec

thomastaylor312:
This PR proposes a new specification section for a "Keyring protocol."
The purpose of this protocol is to enable easier key distribution for the
80% use case. Instead of being something intrinsic to the bindle protocol,
I believe having this as a separate protocol that any server (including existing
key servers) could implement will make it much more flexible.

If this spec change is accepted, I will implement it in a followup PR

### DIFF
--- a/docs/keyring-protocol-spec.md
+++ b/docs/keyring-protocol-spec.md
@@ -1,0 +1,43 @@
+# Keyring Protocol Specification
+
+This specification describes an HTTPS protocol for fetching public key entries for use in Bindle
+keyrings. It is _not required_ to distribute keys in this manner, but it does allow for a convenient
+way to distribute well known keys.
+
+## Path and Protocols
+
+Implementations MUST use HTTP secured with TLS (i.e. HTTPS) in order to guarantee that the received
+data has not been tampered with. Implementations are not required to only host this protocol as part
+of a single server (e.g. they can serve other endpoints besides the keys).
+
+It is RECOMMENDED that implementations serve this protocol on the `/bindle-keys` path, mounted onto
+any path prefixes (e.g. `/api/v1/bindle-keys`).
+
+## Verbs and Parameters
+
+Implementations MUST support the `GET` HTTP verb and MAY support `HEAD`. All other verbs MUST NOT be
+used.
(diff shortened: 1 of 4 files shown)


docs/keyring-protocol-spec.md
```diff
@@ -0,0 +1,39 @@
...
+## Verbs and Parameters
+
+Implementations MUST support the `GET` HTTP verb and MAY support `HEAD`. All other verbs MUST NOT be
+used.
+
+Implementations MUST support the following query string parameters. Other parameters MAY be added by
+implementations, but they are considered non standard. If no query string parameters are specified,
+implementations SHOULD return all available keys.
+
+- `roles`: A comma delimited list of roles to filter on. All keys with these roles will be returned.
```
thomastaylor312, line 25: Please note that this is being kept simple on purpose for now. The types of parameters and filtering could be fairly complex in the future

vdice (commented):
I wonder if we could use more doc additions/examples illustrating the setup of keys for bindle clients and servers.  For example, I'm a bindle client and I know the url to a bindle server and can authn with it.  How do I give the bindle server my public key(s) and how do I get the server public key(s)?  When the server and client are running on different hosts, neither has access to each other's keyrings, etc.

Oh, also, we should add the flow around signing invoices (`bindle sign-invoice`) and pushing bindles w/ signed invoices to the server.  Though perhaps this functionality has been around for some time... in that case, a separate docs issue would be appropriate.

thomastaylor312:
Yeah I think the scope of this spec change is purely:

> how do I get the server public key(s)

Other key distribution methods are specifically excluded from this for now, but would definitely be a welcome addition.

The other suggestion about docs is definitely valid, but not in scope for this PR IMO. We should open an issue and update the docs in another PR

technosophos (approved):

thomastaylor312:
@radu-matei and/or @fibonacci1729 could I get another review on this please? Since this is a spec change, I want to make sure at least one more maintainer reviews this

fibonacci1729 (approved):

PR #298, opened 2022-02-03
ref(cli): Makes global bindle URL less annoying to work with

### LINKED ISSUE #111: Not all `bindle` commands require globally required Bindle URL
Not all of the `bindle` commands require a Bindle server URL. However, it appears that Clap requires that either the env var is set or the flag is explicitly passed.

thomastaylor312:
I just went with the simple workaround of providing a default. It results in
the least amount of code changes and is easy to override with your own value

Closes #111

### DIFF
--- a/bin/client/opts.rs
+++ b/bin/client/opts.rs
@@ -16,7 +16,8 @@ pub struct Opts {
         short = 's',
         long = "server",
         env = "BINDLE_URL",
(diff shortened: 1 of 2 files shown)


vdice:
I'm in support of this; another alternative (more code) would be to limit the restriction on supplying the bindle URL to just those commands that actually use it.  It currently seems to be required for plenty of commands that don't utilize its value.

radu-matei (approved):

PR #300, opened 2022-02-04
feat(*): Enforce signatures on server

### LINKED ISSUE #106: Server needs to validate `creator` signature on invoice receipt
When the server receives a new invoice, it MUST verify that the invoice is signed, that the signature is valid, and that the signing key used is already in the server's keyring. Any failure on this MUST lead to rejecting the bindle.

thomastaylor312:
This completes the change to full enforcement of signing and verification on
both client and server. This PR specifically is for the server. It will no
longer accept bindles that do not have a signature. All signature rules
are enforced according to the verification strategy. Please note that most
of the changes in this PR were to update the tests, which was a good exercise.

This is a breaking behavior change that should be well-documented at the next release.

Closes #106


vdice (commented):
Tested this via a local bindle server instance; verification of signed invoices (and rejection of un-signed invoices) 
seemed to work as expected.  I'll leave the code review details up to the resident Rustaceans 🦀

technosophos (approved):

PR #301, opened 2022-02-04
fix(file): Adds error log when IO errors occur on invoice/parcel create

### LINKED ISSUE #297: Bindle server doesn't log error when it fails writing the bindle to disk
When the bindle process doesn't have write permissions to the bindle directory it silently fails.

To test you can create a temp directory as root.  Pushing a bindle will return a 500 error but the server doesn't log an error. 

```
bindle-server --unauthenticated  -d $(sudo mktemp -d)
```

thomastaylor312:
This will now log an error when the bindle server initially tries to create
directories for invoices or parcels. I checked the embedded provider and these
kind of errors are caught on creation

Closes #297


adamreese (approved):
Manually tested.  Thanks!

PR #309, opened 2022-02-23
feat(cli): add `bindle clean` subcommand to clear the local cache

### LINKED ISSUE #280: Add a `bindle clean` subcommand
Right now the bindle cli uses the `DumbCache` to store files locally. There are many reasons to want to clear this cache, particularly if you are testing bindle locally and nuke your bindle server's store of data. 

We should add a `clean` subcommand to the bindle CLI that removes all cached bindles

VishnuJin:
fixes #280 
This PR adds `clean` subcommand to `bindle cli` to clear the DumpCache

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -145,7 +145,7 @@ async fn run() -> std::result::Result<(), ClientError> {
         .build(&opts.server_url, token, keyring)?;
 
     let local = bindle::provider::file::FileProvider::new(
-        bindle_dir.clone(),
+        &bindle_dir,
         bindle::search::NoopEngine::default(),
     )
     .await;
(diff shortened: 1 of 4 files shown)


bin/client/main.rs
```diff
@@ -433,6 +433,10 @@ async fn run() -> std::result::Result<(), ClientError> {
                 }
             }
         }
+        SubCommand::Clean(_clean_opts) => {
+            // TODO: Clean up cached bindles
```
thomastaylor312, line 437: You missing a commit here?

thomastaylor312: Oh, just saw you move it to draft. Never mind

VishnuJin: @thomastaylor312 sorry Im working on it, this is still in draft

VishnuJin:
hi @thomastaylor312 some brief hints to proceed further on this implementation would be helpful !!

- should `clean` have any args with it
- should we delete everything in the cache dir

thomastaylor312:
I think for this first implementation we don't need any extra flags (other than the top level `bindle-dir` which is included by default). Deleting all the things in that cache dir is also just fine for now!

bin/client/main.rs
```diff
@@ -433,6 +433,19 @@ async fn run() -> std::result::Result<(), ClientError> {
                 }
             }
         }
+        SubCommand::Clean(_clean_opts) => {
+            // Cleans up the local bindles directory.
+            println!("Cleaning up bindles in {:?}...", bindle_dir);
```
thomastaylor312, line 438: Nit: It is generally better to just use `{}` and then use the display implementation of `PathBuf`: `bindle_dir.display()`

bin/client/main.rs
```diff
@@ -145,7 +145,7 @@ async fn run() -> std::result::Result<(), ClientError> {
         .build(&opts.server_url, token, keyring)?;
 
     let local = bindle::provider::file::FileProvider::new(
-        bindle_dir,
+        bindle_dir.clone(),
```
thomastaylor312, line 148: Nit: I think you can just pass a borrow here actually

thomastaylor312 (approved):
Looking good! Just a few optional nits! I'll give it a day in case you'd like to address them and then I'll merge!

bin/client/main.rs
VishnuJin: fixed

bin/client/main.rs
VishnuJin: fixed

PR #314, opened 2022-02-28
fix(cli)-minor-replace the hardcoded bad request error message

### LINKED ISSUE #277: Better error on Bad Request
Right now we have a hardcoded error in the client for when a server returns a `400` status code:

> The request could not be handled by the server. Verify your Bindle server URL

This can be very misleading for cases like trying to create a parcel that doesn't exist in the invoice (which returns a `400`). This error should be updated to use the returned error message.

VishnuJin:
fixes #277


thomastaylor312 (approved):
Awesome, thank you!

PR #316, opened 2022-02-28
DO NOT COMMIT: Add support for mTLS auth

### LINKED ISSUE #312: mTLS auth
To support m2m auth for e.g. CI/CD, add support for client certificate authentication. This would include a config option for specifying a CA certificate which would be used to validate client certificates.

lann:
Fixes #312

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["server", "client", "caching", "test-tools"]
-server = ["warp", "openid"]
+server = ["warp", "openid", "x509-parser"]
 client = ["mime_guess", "dirs", "time"]
 caching = []
 test-tools = []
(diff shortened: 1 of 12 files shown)


lann:
As written this depends on https://github.com/seanmonstar/warp/pull/942 being merged. Another possible approach would be to extract client cert with a Tower layer, though from a brief review that looks like significantly more work unless I am missing an existing crate.

Cargo.toml
```diff
@@ -48,7 +48,7 @@ semver = { version = "1.0.4", features = ["serde"] }
 tokio = { version = "1.11.0", features = ["full"] }
 tokio-util = { version = "0.6.8", features = ["io"] }
 tokio-stream = { version = "0.1.7", features = ["fs"] }
-warp = { version = "0.3", features = ["tls"], optional = true }
+warp = { path = "../../src/warp", features = ["tls"], optional = true }
```
radu-matei, line 51: Would you mind replacing this with a git revision to your fix instead, please?

PR #317, opened 2022-02-28
spec: Require clients follow redirects from parcel GETs

lann:
This gives hosts flexibility in serving parcels from alternative services (CDN, IPFS, S3, etc).

### DIFF
--- a/docs/protocol-spec.md
+++ b/docs/protocol-spec.md
@@ -12,7 +12,7 @@ HTTP Endpoints:
 - `/_i`
     - `POST`: Create a new bindle. If all of the parcels specified in the bindle exist, a 201 status will be returned. If 1 or more of the parcels are missing, a 202 status will be returned with a reference to the missing parcels
 - `/_i/{bindle-name}@{parcel-id}`: The path to a Bindle name and parcel ID, where `{parcel-id}` is an exact SHA of a parcel and `{bindle-name}` follows the same rules as outlined above. Parcels can only be accessed if the client has the proper permissions to access the given bindle and, as such, cannot be accessed directly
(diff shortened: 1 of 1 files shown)


lann:
I _think_ the bindle client already follows redirects (that seems to be the default reqwest behavior), but I haven't tested.

docs/protocol-spec.md
```diff
@@ -275,4 +275,4 @@ See the [Invoice Specification](invoice-spec.md) for a description of the `[bind
 
 ### Ordering of Results
 
-The specification does not provide any guidance on ordering of search results. It is a desirable, but not required, property of search results that under the same circumstances, two identical queries return identical results, including identical ordering.
\ No newline at end of file
+The specification does not provide any guidance on ordering of search results. It is a desirable, but not required, property of search results that under the same circumstances, two identical queries return identical results, including identical ordering.
```
lann, line 278: Github editor did this :shrug:

itowlson:
> I think the bindle client already follows redirects

s/client/Rust client/ *grin* - we'd need to check the .NET and JavaScript clients too (and there might now be a Go client as well).

thomastaylor312:
@itowlson Fun fact: The Go client existed before the .NET and JS clients

thomastaylor312 (approved):
This sounds smart to me. I've already had various conversations with people about this before, so it is a good idea to enshrine it in the spec

thomastaylor312:
@technosophos or @radu-matei @itowlson I'll let one of you do a final review and merged. Figured we need 2 reviewers at least since this is a spec change

radu-matei (approved):

Issue #318, opened 2022-03-02
proposal: Update invoice signature spec


lann:
Following up on #292, I'd like to change the signature spec from signing data _derived_ from the invoice file to signing the invoice file itself. This would require a few related changes:

* The signature(s) would not be stored in the `invoice.toml` itself, but as part of a separate structure (e.g. `invoice_meta.toml`; bikeshedding welcome)
* The server would need to store the bytes of the `invoice.toml` rather than just the semantic data
* While the protocol could still return other invoice formats, only TOML would be usable for verification
* [anything I missed?]

**Disadvantages**
* Signature-verifying Bindle implementations must have a TOML implementation
* Some server-side storage overhead from storing serialized TOML
* Two data structures (files) for invoice data instead of one

**Advantages**
* Avoids any canonicalization attack problems with verification
* Signing and verification is simplified
* Easier to reason about backward compatibility in invoice format

lann:
One option for the invoice signatures file would be to make it a "TOML-compatible" file format, reusing the current spec's `[[signature]]` blocks and appending the signature itself to the end, e.g. something like (adapting from https://github.com/deislabs/bindle/blob/main/docs/signing-spec.md#reading-signatures-as-provenance):

Signature payload (TOML):
```
by = "Matt Butcher <matt.butcher@example.com>"
key = "1c44..."
role = "creator"
at = 1611960337
```

Signature serialization (TOMLish-but-not):
`[[signature]]\n<payload>\nsignature = "<hex(sign(payload))>"`

lann:
I hate to say it, but this would be simplified by using JSON...

fibonacci1729:
@thomastaylor312 Would love your thoughts here!

lann:
Another proposal for signatures, this time more TOMLy:

```toml
[[signature]]
meta = """
by = "Matt Butcher <matt.butcher@example.com>"
key = "1c44..."
role = "creator"
at = 1611960337
"""
signature = "<hex(sign('BINDLE-SIGNATURE-V1' || sha256(invoice) || meta))>"
```

thomastaylor312:
Just getting round to this in #322. I'll drop all comment there

Issue #292, opened 2022-02-01
Improving the Cryptographic Integerity of a Canonical Bindle Invoice


fibonacci1729:
# The Problem

**TL;DR**: Include all relevant semantic information in the canonical invoice when computing the merkle root to protect the signer's intent and preserve the semantic integrity of the stored invoice.

Currently, from the bindle spec, the merkle root of an invoice is computed by hashing the concatenation of the following pieces of data together in a line-separated (`\n`) UTF-8 string:`by`, `name`, `version`, `role`, `at` and the `label.sha256` of each parcel.

E.g:

```
Gnome Grumpsky <gnome.grumpsky@example.com>
mybindle
0.1.0
creator
1611960337
~
e1706ab0a39ac88094b6d54a3f5cdba41fe5a901
098fa798779ac88094b6d54a3f5cdba41fe5a901
5b992e90b71d5fadab3cd3777230ef370df75f5b
```

What is not included in this construction is the integrity of the metadata that is semantically relevant (i.e. `groups`, `features`, `conditions`, `annotations`, and parcel metadata).

A possible attack vector arises for injecting/deselecting parcels from a `bindle` whereby an attacker that compromises a `Bindle` can change a `bindle`'s invoice to augment how parcels are selected without corrupting the `bindle`'s merkle root.

To preserve the cryptographic integrity of the invoice, all relevant semantic information should be included in the canonical invoice before computing the merkle root digest. 

# Proposed Solution

**TL;DR** When creating the canonical invoice of a `bindle`, include all semantically relevant metadata when computing the merkle root digest.

As currently defined by the spec, the canonical invoice of a `bindle` is structured as the concatenation of:

```
<by>
<name>
<version>
<role>
<timestamp>
~
<parcel digest 1>
...
<parcel digest N>
```

Omitting top-level metadata `groups` and `annotations`, as well as the following parcel metadata: `annotations`, `conditions`, `features`,  `media-type`, `name`, and `origin`.

To guarantee integrity of an invoice, we should include these omitted pieces of metadata.

I.e,

```
<by>
<name>
<version>
<role>
<timestamp>
*<annotations>
*<groups>
~
e1706ab0a39ac88094b6d54a3f5cdba41fe5a901
*<annotations> 
*<conditions> 
*<media-type> 
*<name>
*<origin>
...
~
098fa798779ac88094b6d54a3f5cdba41fe5a901
*<annotations>
*<conditions>
*<media-type>
*<name>
*<origin>
```

[...]

PR #322, opened 2022-03-02
Invoice signing spec update

lann:
re #318

### DIFF
--- a/docs/bindle-spec.md
+++ b/docs/bindle-spec.md
@@ -2,18 +2,19 @@
 
 The purpose of the Bindle specification is to describe a format for transporting collections of (potentially large) related files in such a way that they can be safely shared and re-used.
 
-A bindle consists of an `invoice.toml` file and one or more `parcel` objects.
+A bindle consists of `invoice.toml` and `signatures.toml` files and one or more `parcel` objects.
 
 The English word "bindle" refers to a collection or package. A popular portrayal of the bindle is a handkerchief-wrapped bundle hanging off of a stick. (The word itself is likely a corruption of the word "bundle"). This specification describes a format called "bindle" together with an API for storage and retrieval of bindles.
 
 ## Bindles and Parcels
 
-A _bindle_ is a named and versioned container that enumerates its contents in an _invoice_. The contents of a bindle are opaque blobs of data described as _parcels_, where each parcel has a _label_ describing it, and a _parcel.dat_ that contains the opaque data of the parcel.
+A _bindle_ is a named and versioned container that enumerates its contents in an _invoice_, which is authenticated by one or more _signatures_. The contents of a bindle are opaque blobs of data described as _parcels_, where each parcel has a _label_ describing it, and a _parcel.dat_ that contains the opaque data of the parcel.
 
 ```
 BINDLE
   |
-  |- invoice
+  |- invoice.toml
+  |- signatures.toml
   |
   |- parcels
        |- parcel 1
@@ -36,7 +37,7 @@ Bindles are good for expressing the relationship between related chunks of data.
 ### Packaging a website.
 Bindle could be used to model a website by storing each HTML page, image, script, and stylesheet a `parcel`, and then creating an `invoice` that listed each parcel.
 
-As the website changed and evolved, the `invoice` would be versioned (e.g. `v1.0.1`, then `v1.0.2`...). When the website is deployed, the deployment target inspects the `invoice` to see which `parcel`s changed. Then it installs just the changed `parcel`s. This can drastically cut down the amount of data that must be moved around with each website deployment
+As the website changed and evolved, the bindle would be versioned (e.g. `v1.0.1`, then `v1.0.2`...). When the website is deployed, the deployment target inspects the `invoice` to see which `parcel`s changed. Then it installs just the changed `parcel`s. This can drastically cut down the amount of data that must be moved around with each website deployment
 
 ### Packaging Programming Language Dependencies
 
--- a/docs/protocol-spec.md
+++ b/docs/protocol-spec.md
@@ -5,8 +5,8 @@ Bindle uses HTTP/2 with TLS as a transport protocol. All bodies and responses ex
 The HTTP endpoints defined above MAY exist as a subpath on a server, or in the server's root. For example, `https://example.com/v1/_i/foo` and `https://example.com/_i/foo` are both legal paths for the specification below. However, `https://example.com/_i/v1/foo` is not (or, rather, it is a legal URI for a package named `v1/foo`).
 
 HTTP Endpoints:
-- `/_i/{bindle-name}`: The path to a bindle's invoice. Note that `{bindle-name}` can be pathy. For example, `/_i/example.com/mybindle/1.2.3` is a valid path to a bindle named `example.com/mybindle/1.2.3`.
-    - `GET`: Get a bindle by name. This returns an invoice object.
+- `/_i/{bindle-name}`: The path to a bindle. Note that `{bindle-name}` can be pathy. For example, `/_i/example.com/mybindle/1.2.3` is a valid path to a bindle named `example.com/mybindle/1.2.3`.
+    - `GET`: Get a bindle by name.
     - `HEAD`: Send just the headers of a GET request
     - `DELETE`: Yank a bindle. This will set the `yank` field on a bindle to `true`. This is the only mutation allowed on a Bindle.
 - `/_i`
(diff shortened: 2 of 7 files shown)


lann:
This still needs attention to yanking.

lann:
I'm sure I didn't catch everything but I think this is pretty close to complete.

docs/reference-spec.md
```diff
@@ -8,9 +8,9 @@ A bindle has a name. For example, if we were to write a `hello world` bindle, it
 
 When a bindle is concretized into an invoice, it MUST have a name and a version. Version MUST be a valid [SemVer 2 version](https://semver.org).
 
-A bindle name can be one or more characters from the set of unicode alphabetic and numeric characters, as well as the characters `_` and `/`.
+A bindle name can be one or more characters from the set of unicode alphabetic and numeric characters, as well as the characters `-`, `.`, `_` and `/`.
```
lann, line 11: Bit of a drive-by update but seems obvious to me; `.` is used in examples right below and `-` is allowed in domain names.

docs/bindle-spec.md
```diff
@@ -36,7 +37,7 @@ Bindles are good for expressing the relationship between related chunks of data.
 ### Packaging a website.
 Bindle could be used to model a website by storing each HTML page, image, script, and stylesheet a `parcel`, and then creating an `invoice` that listed each parcel.
 
-As the website changed and evolved, the `invoice` would be versioned (e.g. `v1.0.1`, then `v1.0.2`...). When the website is deployed, the deployment target inspects the `invoice` to see which `parcel`s changed. Then it installs just the changed `parcel`s. This can drastically cut down the amount of data that must be moved around with each website deployment
+As the website changed and evolved, the bindle would be versioned (e.g. `v1.0.1`, then `v1.0.2`...). When the website is deployed, the deployment target inspects the `invoice` to see which `parcel`s changed. Then it installs just the changed `parcel`s. This can drastically cut down the amount of data that must be moved around with each website deployment
```
thomastaylor312, line 40: While you are here, could you update these version examples to drop the `v`?

docs/protocol-spec.md
```diff
@@ -5,8 +5,8 @@ Bindle uses HTTP/2 with TLS as a transport protocol. All bodies and responses ex
 The HTTP endpoints defined above MAY exist as a subpath on a server, or in the server's root. For example, `https://example.com/v1/_i/foo` and `https://example.com/_i/foo` are both legal paths for the specification below. However, `https://example.com/_i/v1/foo` is not (or, rather, it is a legal URI for a package named `v1/foo`).
 
 HTTP Endpoints:
-- `/_i/{bindle-name}`: The path to a bindle's invoice. Note that `{bindle-name}` can be pathy. For example, `/_i/example.com/mybindle/1.2.3` is a valid path to a bindle named `example.com/mybindle/1.2.3`.
-    - `GET`: Get a bindle by name. This returns an invoice object.
+- `/_i/{bindle-name}`: The path to a bindle. Note that `{bindle-name}` can be pathy. For example, `/_i/example.com/mybindle/1.2.3` is a valid path to a bindle named `example.com/mybindle/1.2.3`.
+    - `GET`: Get a bindle by name.
```
thomastaylor312, line 9: "Get a bindle object by name" <- Important to connect it with the new section below

docs/protocol-spec.md
```diff
@@ -26,6 +26,55 @@ HTTP Endpoints:
 
 While bindle names MAY be hierarchical, neither the `_i` nor the `_p` endpoints support listing the contents of a URI. This constraint is for both scalability and security reasons. To list available bindles, agents MUST use the `_q` endpoint if implemented. In absence of the `_q` endpoint, this specification does not support any way to list available bindles. However, implementations MAY support alternative endpoints, provided that the URI for those endpoints does not begin with the `_` character.
 
+## Bindle object
+
+Normally, a Bindle object consists of an `invoice` string (with the exact contents signed by the bindle creator) and one or more `signature` blocks:
```
thomastaylor312, line 31: Do we need to specify what kind of data format the `invoice` string is? Should it be the same format as the requested bindle object or should it always be toml?

docs/protocol-spec.md
```diff
@@ -89,7 +138,6 @@ A bindle that is marked `yanked = true` MUST be treated according to the followi
 - It MUST NOT be accepted by a `POST` operation
 - The `DELETE` operation is a no-up on a yanked Bindle
 - A `GET` request should only be fulfilled if the `yanked=true` query parameter is set. In any other case, it should mark it as "access denied"
-    - If `yanked=true` in the query string, the server SHOULD serve the bindle unaltered, including the `invoice.toml`'s `yanked = true` attribute.
```
thomastaylor312, line 92: Any specific reason for removing this functionality?

docs/signing-spec.md
```diff
@@ -579,21 +522,11 @@ However, there are a few non-normative recommendations:
...
 If a proxy or approver key is compromised, a host SHOULD remove all signature blocks made using that key. If a compromise's window of time is known, a host MAY choose instead to remove that signature from only the packages signed during that window.
 
 If at any point the number of trusted signatures on a package is less than one, that package should be deemed untrusted, and appropriate measures taken (including potentially deleting that bindle from the repository). Likewise, if the only trusted key is a proxy key, this may be considered insufficient to warrant full trust, and similar measures may be taken. This security exception is the only case in this specification were outright deletion of a package is condoned.
 
 In all of these cases, note that timestamps used inside of signature blocks from compromised keys should not be trusted.
-Timestamps from the system or from other signatures may serve as better points of trust.
-
-## Questions
-
-### Why Don't You Just Hash The Document?
```
thomastaylor312, line 593: I don't think I feel comfortable just hashing the entire contents of an invoice for many of the reasons discussed elsewhere and the ones listed in this section. There is no guarantee that serialization will be the same among various TOML or JSON serializers in different languages (and possibly even within the same language). Based on what I see in the spec update here, it appears that we attempt to solve this problem by just storing and returning the raw invoice as given to the server when it is created. I feel like there are so many ways this could come back to bite a user, particularly if they are using tooling to introspect a bindle. 

A second, somewhat smaller, problem (probably could be considered a subcategory of the previous one) is with large invoices. Let's say that I have an invoice with several thousand parcels (yes, I know this is currently hypothetical, but I guarantee you _someone_ will try to use it this way). Now if you want to have a copy of the invoice as an actual data object (obtained by parsing the raw version from toml into an object) so you can iterate over it, you now have a huge array of bytes in memory for the raw data and a _second_ version in memory as the parsed object. To be clear, I am less worried about this second problem compared to the first, but I think it illustrates at least one other way this could be painting us into a corner.

I do like the idea of separating signatures and invoices if we can, I still prefer what is described in #292 as a better signing methodology. But I could also just be overthinking this. Any thoughts?

thomastaylor312: Another problem here (which you call out in #318), is that it requires storing the raw bytes of the invoice. This will cause a lot of friction for those people who want to implement a `Provider` that stores things in something like SQL database. We don't have any of those yet, but I consider it increasingly likely as people continue to use bindle

docs/protocol-spec.md
thomastaylor312: After rereading the original issue, I think this is meant to always be TOML. If we choose to go this route of signing, that detail should be specified in the spec

docs/protocol-spec.md
lann: No functionality change here. This specific line is just redundant in this proposal; the invoice is _always_ immutable.

docs/signing-spec.md
lann: >There is no guarantee that serialization will be the same among various TOML or JSON serializers

This is _exactly_ why you want to sign the serialized invoice; the signature(s) can be verified before parsing it at all, so an attacker cannot modify the contents of the invoice and trick you into accepting an existing signature for it.

The existing signature spec has at least one (not obviously exploitable) weakness in this vein: the signature payload uses `\n`-separated fields but the `by` field does not strictly prohibit `\n` in its value, allowing for potential parsing confusion. This specific weakness would be easy enough to resolve, but more importantly this kind of subtle analysis needs to be done every time the spec is updated. Signing the entire document avoids a whole class of similar problem entirely.

lann: > A second, somewhat smaller, problem (probably could be considered a subcategory of the previous one) is with large invoices

I had considered adding a new endpoint in this proposal, e.g.

- `GET /_i/{bindle-name}/invoice`: Get the original bindle invoice as signed by its creator. Clients must follow HTTP redirects for this endpoint.

Along with something like a `include_invoice=false` option on the bindle object GET this would allow the raw invoice to be served from the same storage as parcels.

I ultimately didn't think it was necessary to spec right now but I think it would be a backward-compatible solution if large invoices became a problem.

> This will cause a lot of friction for those people who want to implement a `Provider` that stores things in something like SQL database

The serialized TOML shouldn't be much bigger than when parsed into database fields (perhaps even _smaller_ if compressed) and most DBMSes have no problem with large text fields. `2n` space is the same as `n` space in CS right? :slightly_smiling_face: 

All that said, these do represent real disadvantages that should be weighed against the security implications.

thomastaylor312: @radu-matei @technosophos I'd love your thoughts on everything in this thread as well

@lann Just one thing I wanted to address from your comments

> The serialized TOML shouldn't be much bigger than when parsed into database fields (perhaps even smaller if compressed) and most DBMSes have no problem with large text fields. 2n space is the same as n space in CS right? 

It isn't as much about the space involved as it is about query-ability. The primary goal of writing a `Provider` backed by a DB is the ability to easily index the fields that you want to be able to search on (particularly for the query endpoint and any future additions to the relationships endpoint). This also could come in handy when checking if a parcel is part of a bindle (i.e. a user accesses a parcel _through_ an invoice because of authz needs) as a DB makes it really easy to check. Obviously the easy way around this would be to store both the raw bytes and the actual fields in the same row, but this does double the amount of space used again and is a bit clunky for people implementing it

That leads to my other thought is that we also double caching needs using this method. If you want to cache invoices, you need to cache the raw bytes _and_ an in memory version of the invoice with the current providers. Right now we have an actual `Invoice` object available in an LRU cache that allows us to query that data or return it. Even in an eventual future where we have caching that is external to the `bindle-server` binary, we are still storing it twice _again_. Yes, you could store just the raw copy and just deserialize every time you need it, but that could come with a performance overhead on a server with high load. 

I don't think these would be entirely blocking concerns, but we do need to be careful and document these tradeoffs well in the FAQ section if we go down this route

lann: Agreed, this is a pain.

We could change the protocol so that the _raw_ invoice is always served with the same type of endpoint as parcels (e.g. `GET /_i/{bindle-name}/invoice`, following redirects) while still serving the _parsed_ invoice fields on other endpoints, which would I think remove some of this provider implementation pain since raw invoices could use the same storage as parcels. The tradeoff here would be an extra "fetch raw invoice" step on the client before verifying signatures.

PR #323, opened 2022-03-16
Make "unknown error" client errors a bit more informative

itowlson:
The current "Unknown error" is very hard to diagnose because it depends on the server providing everything you need in the message.  This is not a good thing to rely on, especially for unfortunate server devs trying to get their stuff working with the existing client, so this PR adds the following information to the "unknown error" message:

* The URL that produced the unexpected response
* The operation being carried out on that URL
* The status code that was returned

A minor change is that it will now say `Unknown error: (no error message in response)` rather than a trailing colon and space `Unknown error: `.  A complete message now looks like `Unknown error response: Create to http://localhost:5000/v1/_i/weather/1.2.3-ivan-20220317112555054@a4d030d845438e530f12cbf85422e82a65327443786e0c9f59fbbf282adde7e3 returned status 201 Created: (no error message in response)`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,7 +14,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.3",
+ "getrandom 0.2.5",
  "once_cell",
  "version_check",
 ]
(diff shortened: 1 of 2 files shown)


itowlson:
@thomastaylor312 do you own the GH actions for this?  I am getting a failure on something to do with large regexes, but my PR does not change anything involving large regexes.

thomastaylor312:
@itowlson Yeah, this is due to a new RUSTSEC vulnerability. It shouldn't effect us but running a `cargo update` should solve the problem

thomastaylor312 (approved):
This is much better! Feel free to merge if you don't want to do the `cargo update` and I'll handle it later

radu-matei (approved):

PR #329, opened 2022-03-31
Remove pinned indexmap dependency

lann:
I couldn't figure out exactly what was originally breaking that needed this workaround, but bindle seems to build fine now without it.

This pinned version was breaking when trying to link with `toml_edit` in spin due to a version constraint incompatibility.


thomastaylor312 (approved):
Glad this somehow resolved itself! Thanks for taking care of it

PR #330, opened 2022-04-06
feat - add label search argument to sign-invoice

### LINKED ISSUE #299: Add label search to `sign-invoice`
Right now the `sign-invoice` command selects the first key it finds with the proper role. The command should be able to select using an exact label (`--label`) or a label match (`--label-match`).

VishnuJin:
fixes #299

this PR adds optional ability to `sign-invoice` with a key that matches the given label

 ### For example
#### `--label-matching` (partial match)
```
❯ bindle sign-invoice -o signed-invoice.toml invoice.toml --label-matching "Vishnu Jin"
Signed invoice.toml with role creator, label as 'Vishnu Jin <vishnujin@outlook.com>' and wrote to signed-invoice.toml
```

#### `--label ` (exact match)
```
❯ bindle sign-invoice -o signed-invoice.toml invoice.toml --label "Vishnu Jin <vishnujin@outlook.com>"
Signed invoice.toml with role creator, label as 'Vishnu Jin <vishnujin@outlook.com>' and wrote to signed-invoice.toml
```

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -11,7 +11,7 @@ use bindle::invoice::signature::{
 };
 use bindle::invoice::Invoice;
 use bindle::provider::ProviderError;
-use bindle::signature::{KeyEntry, KeyRingLoader, KeyRingSaver};
+use bindle::signature::{KeyEntry, KeyRingLoader, KeyRingSaver, LabelMatch};
 use bindle::standalone::{StandaloneRead, StandaloneWrite};
 use bindle::SignatureError;
 use bindle::{
@@ -246,8 +246,17 @@ async fn run() -> std::result::Result<(), ClientError> {
                 None => ensure_config_dir().await?.join("secret_keys.toml"),
             };
 
+            let match_type = match (sign_opts.label, sign_opts.label_matching) {
+                (Some(label), None) => Some(LabelMatch::FullMatch(label)),
+                (None, Some(label_matching)) => Some(LabelMatch::PartialMatch(label_matching)),
+                (None, None) => None,
+                _ => {
+                    unreachable!("both label and label-matching cannot be present at the same time")
+                }
+            };
+
             // Signing key
-            let key = first_matching_key(keyfile, &role).await?;
+            let key = first_matching_key(keyfile, &role, match_type.as_ref()).await?;
 
             // Load the invoice and sign it.
             let mut inv: Invoice = bindle::client::load::toml(sign_opts.invoice.as_str()).await?;
@@ -259,8 +268,8 @@ async fn run() -> std::result::Result<(), ClientError> {
                 .unwrap_or_else(|| format!("./invoice-{}.toml", inv.canonical_name()));
 
             println!(
-                "Signed {} with role {} and wrote to {}",
-                sign_opts.invoice, role, outfile
+                "Signed {} with role as '{}', label as '{}' and wrote to {}",
+                sign_opts.invoice, role, key.label, outfile
             );
             tokio::fs::write(outfile, toml::to_string(&inv)?).await?;
         }
@@ -646,12 +655,16 @@ async fn ensure_config_dir() -> Result<PathBuf> {
     Ok(dir)
 }
 
-async fn first_matching_key(fpath: PathBuf, role: &SignatureRole) -> Result<SecretKeyEntry> {
+async fn first_matching_key(
+    fpath: PathBuf,
+    role: &SignatureRole,
+    label_match: Option<&LabelMatch>,
+) -> Result<SecretKeyEntry> {
     let keys = SecretKeyFile::load_file(&fpath).await.map_err(|e| {
         ClientError::Other(format!("Error loading file {}: {}", fpath.display(), e))
     })?;
 
-    keys.get_first_matching(role)
+    keys.get_first_matching(role, label_match)
         .map(|k| k.to_owned())
         .ok_or_else(|| ClientError::Other("No satisfactory key found".to_owned()))
 }
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -324,7 +324,7 @@ async fn run() -> std::result::Result<(), ClientError> {
                         .await
                         .map_err(|e| ClientError::Other(e.to_string()))?;
 
-                    let matches: Vec<KeyEntry> = match print_key_opts.label {
+                    let matches: Vec<KeyEntry> = match print_key_opts.label_matching {
                         Some(name) => keyfile
                             .key
                             .iter()
--- a/bin/client/opts.rs
+++ b/bin/client/opts.rs
@@ -419,12 +419,11 @@ pub struct PrintKey {
     )]
     pub secret_file: Option<PathBuf>,
     #[clap(
-        short = 'l',
-        long = "label",
-        value_name = "LABEL",
-        help = "The label to search for. If supplied, this will return each key that contains this string in its label. For example, '--label=ample' will match 'label: Examples'."
+        short = 'm',
+        long = "label-matching",
+        help = "selects the keys that (partially) matches the given label. If supplied, this will return each key that contains this string in its label. For example, '--label=ample' will match 'label: Examples'."
     )]
-    pub label: Option<String>,
+    pub label_matching: Option<String>,
 }
 
 #[derive(Parser)]
--- a/bin/client/opts.rs
+++ b/bin/client/opts.rs
@@ -448,6 +448,19 @@ pub struct SignInvoice {
     )]
     pub role: Option<String>,
     #[clap(
+        short = 'l',
+        long = "label",
+        conflicts_with = "label-matching",
+        help = "selects the key with the exact given label"
+    )]
+    pub label: Option<String>,
+    #[clap(
+        short = 'm',
+        long = "label-matching",
+        help = "selects the key that (partially) matches the given label. If supplied, this will sign with the key that contains this string in its label. For example, '--label=ample' will match 'label: Examples'."
+    )]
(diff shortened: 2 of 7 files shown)


thomastaylor312 (changes requested):
2 minor changes here.

Also, it would be great if we could have a `--label` that does an exact match, but that is optional

bin/client/opts.rs
```diff
@@ -447,6 +447,12 @@ pub struct SignInvoice {
         help = "the role to sign with. Values are: c[reator], a[pprover], h[ost], p[roxy]. If no role is specified, 'creator' is used"
     )]
     pub role: Option<String>,
+    #[clap(
```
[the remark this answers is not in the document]
VishnuJin: @thomastaylor312 yeah I agree, but I saw `--label` used for matching in `keys print` here, probably we may need to change that as well
https://github.com/deislabs/bindle/blob/f7e31ccc5587c7444ef73bba702ec082f37cf345/bin/client/opts.rs#L421-L427

VishnuJin:
@thomastaylor312  can we use the `-l` as short for `--label` and `-m` as short for `--label-match` ?

bin/client/opts.rs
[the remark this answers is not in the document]
thomastaylor312: Yeah, that is a good call out. We should make those match

bin/client/main.rs
```diff
@@ -246,8 +246,21 @@ async fn run() -> std::result::Result<(), ClientError> {
...
+
             // Signing key
-            let key = first_matching_key(keyfile, &role).await?;
+            let key = label
+                .as_ref()
+                .map_or_else(
+                    || keys.get_first_matching(&role),
+                    |label| keys.get_first_label_matching(&role, &label),
+                )
+                .map(|k| k.to_owned())
```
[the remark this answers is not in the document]
VishnuJin: done

src/invoice/signature.rs
```diff
@@ -419,6 +424,15 @@ impl SecretKeyStorage for SecretKeyFile {
...
         self.key.iter().find(|k| k.roles.contains(role))
     }
+    fn get_first_label_matching(
+        &self,
+        role: &SignatureRole,
+        label: &str,
+    ) -> Option<&SecretKeyEntry> {
+        self.key
+            .iter()
+            .find(|k| k.label.contains(label) & k.roles.contains(role))
```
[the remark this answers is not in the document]
VishnuJin: my bad, modified

bin/client/opts.rs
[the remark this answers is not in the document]
VishnuJin: I have changed the `--label` to `--label-matching`

bin/client/main.rs
```diff
@@ -246,8 +246,21 @@ async fn run() -> std::result::Result<(), ClientError> {
...
+
             // Signing key
-            let key = first_matching_key(keyfile, &role).await?;
+            let key = label
+                .as_ref()
+                .map_or_else(
+                    || keys.get_first_matching(&role),
+                    |label| keys.get_first_label_matching(&role, &label, sign_opts.label.is_some()),
+                )
+                .map(|k| k)
```
thomastaylor312, line 262: Nit: This map does nothing and can be removed

thomastaylor312 (approved):
Awesome! Thanks for going back to make things match. I'll leave this open for a bit in case you want to handle the nit and will merge tomorrow!

src/invoice/signature.rs
```diff
@@ -370,7 +370,11 @@ pub trait SecretKeyStorage {
     /// If no key is found, this will return a None.
     /// In general, if multiple keys match, the implementation chooses the "best fit"
     /// and returns that key.
-    fn get_first_matching(&self, role: &SignatureRole) -> Option<&SecretKeyEntry>;
+    fn get_first_matching(
```
VishnuJin, line 373: @thomastaylor312 Just thought it would be better to just have one method in the trait rather than having 2 methods that mostly does the same

VishnuJin:
@thomastaylor312 I have just modified the approach here, if you like this approach I will remove the earlier commits !! let me know your thoughts

bin/client/main.rs
```diff
@@ -246,8 +246,21 @@ async fn run() -> std::result::Result<(), ClientError> {
                 None => ensure_config_dir().await?.join("secret_keys.toml"),
             };
 
-            // Signing key
-            let key = first_matching_key(keyfile, &role).await?;
+            let keys = SecretKeyFile::load_file(&keyfile).await.map_err(|e| {
+                ClientError::Other(format!("Error loading file {}: {}", keyfile.display(), e))
+            })?;
+
+            let match_type = if sign_opts.label.is_some() {
```
thomastaylor312, line 253: I really like the enum way of doing this! 

One nit here: Probably better to do an `if let Some(label) = sign_opts.label.as_ref()` instead

src/invoice/signature.rs
```diff
@@ -415,9 +419,32 @@ impl SecretKeyFile {
...
+impl<'a> Default for LabelMatch<'a> {
+    fn default() -> Self {
+        Self::None
+    }
+}
+
 impl SecretKeyStorage for SecretKeyFile {
-    fn get_first_matching(&self, role: &SignatureRole) -> Option<&SecretKeyEntry> {
-        self.key.iter().find(|k| k.roles.contains(role))
+    fn get_first_matching(
```
thomastaylor312, line 435: Yep, this is much cleaner and has the advantage of being a single function rather than a separate one

src/invoice/signature.rs
```diff
@@ -415,9 +419,32 @@ impl SecretKeyFile {
...
+        &self,
+        role: &SignatureRole,
+        label_match: LabelMatch,
```
thomastaylor312, line 438: Would it be better if we did `Option<LabelMatch>` and drop the `LabelMatch::None`? That might be a bit cleaner and makes it so people don't have to import `LabelMatch` if they aren't using it

thomastaylor312 (commented):
I actually really like this way of doing it! Just a few comments for discussion

src/invoice/signature.rs
VishnuJin: yeah, I thought its good to have a `Default` implementation that can be used in test files etc rather than just using `None`
but let me know if I can change this to `Option<LabelMatch >` !!

bin/client/main.rs
VishnuJin: Used Match now, as we are matching based on 2 expressions, using `if let` I felt is a bit strange !!
https://github.com/VishnuJin/bindle/blob/e9e0100b3cca0f5301649f6e5d213fd9aab54f56/bin/client/main.rs#L253-L260

src/invoice/signature.rs
VishnuJin: I changed it to Option<LabelMatch> and removed `LabelMatch::None`

VishnuJin:
@thomastaylor312 this is ready for review now

src/invoice/signature.rs
```diff
@@ -414,10 +418,32 @@ impl SecretKeyFile {
         Ok(())
     }
 }
+/// This enumerates the select criteria of the key based on the given Label
+pub enum LabelMatch {
```
thomastaylor312, line 422: Sorry for another comment here. This is entirely optional, but just wanted your opinion as this API will be used relatively often and I was wondering if this could be cleaner:

```rust
pub enum LabelMatch<'a> {
    /// Key will be selected with an exact match of the given label.
    /// Matches are case sensitive
    FullMatch(&'a str),
    /// Key will be selected with a partial match of the given label,
    /// Matches are case insensitive.
    ///
    /// For example: "pika" will match "Pikachu" and "puff" will match "Jigglypuff"
    PartialMatch(&'a str),
}
```

This would make the enum cheap to construct and no need to copy the `String`. It also would avoid having an `Option<&LabelMatch>` in a function signature.

We can also just merge and come back to this later, but if you do have an opinion, I'd love to hear it!

bin/client/main.rs
```diff
@@ -246,8 +246,17 @@ async fn run() -> std::result::Result<(), ClientError> {
                 None => ensure_config_dir().await?.join("secret_keys.toml"),
             };
 
+            let match_type = match (sign_opts.label, sign_opts.label_matching) {
+                (Some(label), None) => Some(LabelMatch::FullMatch(label)),
+                (None, Some(label_matching)) => Some(LabelMatch::PartialMatch(label_matching)),
+                (None, None) => None,
+                _ => {
+                    unreachable!("both label and label-matching cannot be present at the same time")
```
thomastaylor312, line 254: Interesting...did the compiler not think you'd exhausted all the enum options?

VishnuJin: `Some(), Some()` combination is still pending right, which is not possible as CLAP will throw error because `label` and `label-matching` conflict each other as we have set it to be

src/invoice/signature.rs
VishnuJin: yeah, I had this with lifetime at first but when having it inside `first_matching_key` function in `main.rs` I had trouble with the life time reference as expected. Since these string are small I thought its okay to not use reference. 
But can I give it another try to have `&'a str` like before

bin/client/main.rs
thomastaylor312: Oh right. That makes total sense

src/invoice/signature.rs
thomastaylor312: Nah, no problem. We'll keep this as is and revisit down the line if we want to try again!

thomastaylor312 (approved):

PR #332, opened 2022-04-14
Fix: head method not allowed for query `_q` path

### LINKED ISSUE #331: GET /v1/_q returns data with 405 Not Allowed
If I run bindle-server with the following parameters:

```
><> bindle-server --unauthenticated
```

Then try to `curl` the endpoint, I receive a 405 Not Allowed:

```
><> curl http://localhost:8080/v1/_q -I
HTTP/1.1 405 Method Not Allowed
content-type: text/plain; charset=utf-8
content-length: 23
date: Thu, 07 Apr 2022 17:04:18 GMT
```

...But I still get data back from the server.

```
><> curl -s http://localhost:8080/v1/_q | tail -n 2
role = "host"
at = 1647369693
```

Is this unintentional behaviour?

VishnuJin:
fixes #331 

#### after fix 
```
❯ curl http://localhost:8080/v1/_q\?q\=mybind -I                            
HTTP/1.1 200 OK
content-type: application/toml
content-length: 804
date: Thu, 14 Apr 2022 20:51:19 GMT
```

### DIFF
--- a/src/server/routes.rs
+++ b/src/server/routes.rs
@@ -121,7 +121,7 @@ pub mod v1 {
             S: Search + Clone + Send + Sync,
         {
             warp::path("_q")
-                .and(warp::get())
+                .and(warp::get().or(warp::head()).unify())
                 .and(warp::query::<crate::QueryOptions>())
                 .and(warp::any().map(move || index.clone()))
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
Thanks again for another great bugfix!

PR #334, opened 2022-04-29
feat(*): Adds bindle-keys endpoint support

thomastaylor312:
This adds support for the new `bindle-keys` endpoint to the bindle server as well as client code for consuming that end point. Note that this has one small breaking change for the `raw` method on the `Client`. 

I also fixed some clippy lints that had snuck past us

### DIFF
--- a/bin/client/main.rs
+++ b/bin/client/main.rs
@@ -440,6 +440,40 @@ async fn run() -> std::result::Result<(), ClientError> {
                         .map_err(|e| ClientError::Other(e.to_string()))?;
                     println!("Wrote key to keyring file at {}", keyring_path.display())
                 }
+                Keys::Fetch(opts) => {
+                    let new_keys = match opts.key_server {
+                        Some(url) if !opts.use_host => {
+                            println!("Fetching host keys from {}", url);
+                            get_host_keys(url).await?
+                        }
+                        _ => {
+                            println!("Fetching host keys from bindle server");
+                            bindle_client.get_host_keys().await?
+                        }
+                    };
+
+                    let mut keyring = keyring_path
+                        .load()
+                        .await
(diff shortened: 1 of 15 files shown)


thomastaylor312:
I am going to be out of town for a bit so no rush on this, but I did want to get it open so people could take a look at it and review.

The reason this is still a WIP is because I haven't written tests for the new CLI command. I will finish that when I get back

bin/client/main.rs
```diff
@@ -440,6 +440,40 @@ async fn run() -> std::result::Result<(), ClientError> {
                         .map_err(|e| ClientError::Other(e.to_string()))?;
                     println!("Wrote key to keyring file at {}", keyring_path.display())
                 }
+                Keys::Fetch(opts) => {
+                    let new_keys = if opts.use_host || opts.key_server.is_none() {
+                        println!("Fetching host keys from bindle server");
+                        bindle_client.get_host_keys().await?
+                    } else {
+                        // SAFETY: Can unwrap because we drop into the logic block above if it was
```
dicej, line 448: Not a big deal, but if you wanted to avoid the `unwrap` (and avoid having to explain why it's okay), you could use something like:

```rust
let new_keys = match &opts.key_server {
    Some(url) if !opts.use_host => {
        println!("Fetching host keys from {}", url);
        get_host_keys(url).await
    }
    _ => {
        println!("Fetching host keys from bindle server");
        bindle_client.get_host_keys().await
    }
}?;
```

thomastaylor312:
Sorry this took longer for me to come back around to. @technosophos or @radu-matei this needs a review!

technosophos (approved):
LGTM. Sorry for being so late. I had thought I LGTMed it when I initially reviewed it.

PR #340, opened 2022-07-05
Replaced reqwest's `default-tls` with `rustls-tls`

brooksmtownsend:
This PR replaces the default TLS feature flag for `reqwest` with `rustls-tls` in order to eliminate the openssl dependency when using the `reqwest` crate. I went through a few iterations but couldn't quite find a way to cleanly keep the default-tls feature by default and then able `rustls-tls` optionally here, but I'm happy to implement that to keep the default behavior of this crate the same if desired. 

My main reason for doing this is so that I can depend on `bindle` without bringing in the openssl requirement which can be difficult to deploy across different versions of Linux

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,7 +14,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.5",
+ "getrandom 0.2.7",
  "once_cell",
  "version_check",
 ]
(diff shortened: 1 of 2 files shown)


bacongobbler:
I know Krustlet uses a feature called `rustls-tls` and `native-tls` to differentiate between the two. Would it be valuable to make this a new feature flag?

https://github.com/krustlet/krustlet/blob/29e8a30cbb7fd99e27de061b81e297f308797653/Cargo.toml#L29-L42

brooksmtownsend:
> I know Krustlet uses a feature called [`rustls-tls` and `native-tls`](https://github.com/krustlet/krustlet/blob/29e8a30cbb7fd99e27de061b81e297f308797653/Cargo.toml#L29-L42) to differentiate between the two, with `native-tls` [being the default](https://github.com/krustlet/krustlet/blob/29e8a30cbb7fd99e27de061b81e297f308797653/Cargo.toml#L28). Would it be valuable to make this a new feature flag?

I'd be happy to make this the way you can select the other feature, it would just require that if you disable default features you would have to specify `native-tls` or `rustls-tls` if you brought in the `server` or `client` feature

brooksmtownsend:
Did some good pairing with @thomastaylor312 and got a good way to make this work with the `?` syntax with features 😄 Should be ready for review / tests

thomastaylor312 (approved):

PR #352, opened 2022-10-26
feat: push docker image

FrankYang0529:
ref: https://github.com/fermyon/installer/issues/63, https://github.com/fermyon/installer/pull/100#issuecomment-1276462821

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -191,3 +191,45 @@ jobs:
       - name: Cargo publish
         run: cargo publish
         shell: bash
+  docker-image:
+    name: Build and push docker images 
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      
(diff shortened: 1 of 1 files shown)


FrankYang0529:
@microsoft-github-policy-service agree [company="default"]

FrankYang0529:
@microsoft-github-policy-service agree

vdice (commented):
Defer to the code owners for final decision, but I'm wondering if we want to push to this repo's GitHub container registry instead (`ghcr.io/deislabs/bindle`)?  We should be able to use the default `$GITHUB_TOKEN` to [push images](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry).

FrankYang0529:
> Defer to the code owners for final decision, but I'm wondering if we want to push to this repo's GitHub container registry instead (`ghcr.io/deislabs/bindle`)? We should be able to use the default `$GITHUB_TOKEN` to [push images](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry).

Yeah, I think it's good to use the default `$GITHUB_TOKEN` and we don't need to maintain docker credentials. Thanks. Updated it!

vdice (approved):
Nice, I appreciate how this enables forks to test/run this same workflow as well 👍

Issue #350, opened 2022-10-18
Bindle server version endpoint


vdice:
Is there a pre-existing endpoint that returns the version that a bindle server is running with?

@FrankYang0529 and I were thinking of potentially adding this info to the `healthz` endpoint... thoughts/input?

PR #353, opened 2022-10-27
feat(healthz): add version

FrankYang0529:
resolve https://github.com/deislabs/bindle/issues/350
ref https://github.com/fermyon/spin/pull/786#issuecomment-1283146177

**Test step**
1. Build.
```
> make build
```
2. Run the server.
```
> ./target/debug/bindle-server --unauthenticated
```
3. Test `/healthz`.
```
> curl -i http://localhost:8080/healthz
HTTP/1.1 200 OK
content-type: application/json
content-length: 38
date: Thu, 27 Oct 2022 11:37:31 GMT

{"status":"OK","version":"0.9.0-rc.1"}
```

### DIFF
--- a/src/server/routes.rs
+++ b/src/server/routes.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use warp::Filter;
 
-use crate::{server::filters, signature::KeyRing};
+use crate::{invoice::HealthResponse, server::filters, signature::KeyRing};
 
 /// A helper function that aggregates all routes into a complete API filter. If you only wish to
 /// serve specific endpoints or versions, you can assemble them with the individual submodules
@@ -22,7 +22,12 @@ where
     Authn: crate::authn::Authenticator + Clone + Send + Sync + 'static,
     Authz: crate::authz::Authorizer + Clone + Send + Sync + 'static,
 {
-    let health = warp::path("healthz").map(|| "OK");
+    let health = warp::path("healthz").map(|| {
+        warp::reply::json(&HealthResponse {
+            status: "OK".to_string(),
+            version: env!("CARGO_PKG_VERSION").to_string(),
+        })
+    });
 
     // Use an Arc to avoid a possibly expensive clone of the keyring on every API call
     let wrapped_keyring = Arc::new(keyring);
--- a/src/invoice/api.rs
+++ b/src/invoice/api.rs
@@ -25,6 +25,12 @@ pub struct MissingParcelsResponse {
     pub missing: Vec<Label>,
 }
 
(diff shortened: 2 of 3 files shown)


vdice (commented):
Nice, this looks good to me.  Ping @thomastaylor312 for a second review.

Cargo.toml
```diff
@@ -27,7 +27,7 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["server", "client", "caching", "test-tools", "native-tls"]
-server = ["warp", "openid", "hyper", "mime", "either", "_common"]
+server = ["clap", "warp", "openid", "hyper", "mime", "either", "_common"]
```
thomastaylor312, line 30: We should not be bringing in `clap` for a server dependency. It is fairly big and we don't want people to import it unless they are actually building the CLI.

src/server/routes.rs
```diff
@@ -22,7 +29,12 @@ where
     Authn: crate::authn::Authenticator + Clone + Send + Sync + 'static,
     Authz: crate::authz::Authorizer + Clone + Send + Sync + 'static,
 {
-    let health = warp::path("healthz").map(|| "OK");
+    let health = warp::path("healthz").map(|| {
+        warp::reply::json(&HealthStatus {
+            status: "OK".to_string(),
+            version: clap::crate_version!().to_string(),
```
thomastaylor312, line 35: This can be swapped out to a non-clap thing by doing `env!("CARGO_PKG_VERSION")`

src/server/routes.rs
```diff
@@ -1,9 +1,16 @@
 use std::sync::Arc;
 
+use serde::{Deserialize, Serialize};
 use warp::Filter;
 
 use crate::{server::filters, signature::KeyRing};
 
+#[derive(Deserialize, Serialize)]
+pub struct HealthStatus {
```
thomastaylor312, line 9: Is this a type that we expect people to read back from the server using a client? If so, would it be better to put into `invoice/api.rs` or another location?

thomastaylor312 (changes requested):

Cargo.toml
FrankYang0529: Removed it. Thanks!

src/server/routes.rs
FrankYang0529: Updated it. Thank you.

src/server/routes.rs
FrankYang0529: Moved it to `invoice/api.rs` and also changed the name to `HealthResponse`.

thomastaylor312 (approved):

PR #360, opened 2022-12-13
fix(release.yaml): add apt-get update in aarch64 build step

vdice:
Fix for error seen recently in the release workflow, eg https://github.com/deislabs/bindle/actions/runs/3679702313/jobs/6224462114

I tested with a fresh/stock Ubuntu 22.04 AMI in AWS.  Reproduced the error and it looks like the `apt-get update` indeed unblocks the apt installs.


thomastaylor312:
Oh I missed you had a PR for this when I commented on another issue. This is looking good!

thomastaylor312 (approved):

PR #366, opened 2023-03-14
fix(*): Updates deps to make sure we avoid native tls

thomastaylor312:
This also fixes the problem with linux builds not working on the previous version of OS. See the comment in the pipeline for more details


thomastaylor312:
I don't think any current listed maintainers are still active on this project, so I am going to go ahead and merge

